PR #44, opened 2014-09-06
Added support for Font::Info family member

zsbzsb:
https://github.com/LaurentGomila/SFML/commit/7caf2e64b6aa40c28367e736592c44976e58170f

If you want I can modify to the PR to actually add another struct type for the Font::Info type, I didn't in this version because I figured it would be overkill for something as a single string. If in the future that info struct ever has more members added it won't be any issue to add another function for that member.


zsbzsb:
I am going to start a fresh PR for this once I update it as the entire thing will basically be redone from the way it was originally implemented.

PR #51, opened 2015-01-01
Added extern "C" inside ifdef __cplusplus to allow including and linking CSFML correctly in C++.

FRex:
This fixes https://github.com/LaurentGomila/CSFML/issues/34

### DIFF
--- a/include/SFML/Config.h
+++ b/include/SFML/Config.h
@@ -34,6 +34,16 @@
 
 
 ////////////////////////////////////////////////////////////
+// Check if we need to mark functions as extern "C"
+////////////////////////////////////////////////////////////
+#ifdef __cplusplus
+    #define CSFML_EXTERN_C extern "C"
+#else
+    #define CSFML_EXTERN_C extern
+#endif
+
+
+////////////////////////////////////////////////////////////
 // Identify the operating system
 ////////////////////////////////////////////////////////////
 #if defined(_WIN32) || defined(__WIN32__)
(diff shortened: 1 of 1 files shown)


LaurentGomila:
Hmm what about

```cpp

#ifdef __cplusplus
    #define EXTERN_C extern "C"
#else
    #define EXTERN_C extern
#endif

#define CSFML_API_EXPORT EXTERN_C __attribute__ ((__visibility__ ("default")))
#define CSFML_API_IMPORT EXTERN_C __attribute__ ((__visibility__ ("default")))

... same for the two other occurences without duplicating the whole logic ...
```

FRex:
There you go.
BTW: a valid (but rare) use case for this is when using CSFML and your own C++ library (via extern "C" interface) from some other language, then you will need to use CSFML in C++ to work with SFML objects that this other language passes to you via your C API.

LaurentGomila:
Thanks.

Just one minor thing: I think you can now use ```CSFML_EXTERN_C``` export macros too (like I did in my code snippet above). It won't make any difference because CSFML will always be compiled with a C++ compiler anyway, but it just looks better.

FRex:
I'd say it looks worse and suggest you can compile CSFML as C or something similar.
It makes no sense to conditionally 'allow' just extern when it can never happen in reality.

LaurentGomila:
Ok...

Thanks for the contribution :)

PR #53, opened 2015-01-06
Updated CSFML's version to 2.2.0.

eXpl0it3r:
Once it's confirmed that everything has been updated code wise, this patch can be used to update the version number.


BlaXpirit:
I think it's time.

zsbzsb:
Yes, nothing else is missing as far as I can see. We are good to go :smile:

LaurentGomila:
Nop

https://github.com/LaurentGomila/CSFML/issues/50#issuecomment-75840719

zsbzsb:
>Nop

Uh, only the touch and sensor API isn't in CSFML, but as I was told this is only part of the iOS and Android implementations (which aren't even officially released yet). So unless we plan on supporting mobile with CSFML (requires a lot more than the sensor/touch API) there is nothing else to do.

LaurentGomila:
Touch and sensors are in the public API of SFML 2.2, no matter how or where they are used. iOS and Android are not the only ones to support sensors and touch screens, even if for now we have no implementation for other OSes. The point is that it's part of the public API, so bindings must follow.

zsbzsb:
https://github.com/LaurentGomila/CSFML/issues/50#issuecomment-76244674

So.... :wink:

eXpl0it3r:
Rebased and applied the soversion decision.

LaurentGomila:
Thanks :+1:

Issue #28, opened 2013-09-17
Add FindSFML.cmake comment to readme.txt


zester:
I took me forever to find a solution to CSFML not being able to find SFML the solution was to copy FindSFML.cmake to my /usr/share/cmake2.8/Modules directory on linux.
 
Could you make that more clear in CSFML's readme for others.

eXpl0it3r:
I guess, this can get closed now.

PR #55, opened 2015-01-06
Updated the readme.txt and added a comment about FindSFML.cmake.

eXpl0it3r:
* Fixed links for the new website
* Changed Laurent's email address to match with the SFML readme
* Added a comment about FindSFML.camke (see #28)


LaurentGomila:
Thanks :+1:

PR #57, opened 2015-02-24
Feature/sound recorder

zsbzsb:
*This is not been tested yet as I am still working on the SFML.NET version*

I am opening this PR for a code review. Since handling strings in a C wrapper is a tricky subject....

The main thing I am looking at is if there is a better way to handle returning the available devices. Since I need something to keep the std::string array alive I made a static vector of them. Also since I also need to return an array of C style strings I made another vector to hold the pointers to the actual strings. If anyone could see a better way to handle this without multiple vectors I am open for suggestions.

### DIFF
--- a/include/SFML/Audio/SoundRecorder.h
+++ b/include/SFML/Audio/SoundRecorder.h
@@ -76,8 +76,10 @@ CSFML_AUDIO_API void sfSoundRecorder_destroy(sfSoundRecorder* soundRecorder);
 /// \param soundRecorder Sound recorder object
 /// \param sampleRate    Desired capture rate, in number of samples per second
 ///
+/// \return True, if start of capture was successful
+///
 ////////////////////////////////////////////////////////////
-CSFML_AUDIO_API void sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate);
+CSFML_AUDIO_API sfBool sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate);
 
 ////////////////////////////////////////////////////////////
 /// \brief Stop the capture of a sound recorder
@@ -131,6 +133,57 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
 ///
 ////////////////////////////////////////////////////////////
 CSFML_AUDIO_API void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval);
+
+////////////////////////////////////////////////////////////
+/// \brief Get a list of the names of all availabe audio capture devices
+///
+/// This function returns an array of strings (null terminated),
+/// containing the names of all availabe audio capture devices.
+/// If no devices are available then NULL is returned.
+///
+/// \param count Pointer to a variable that will be filled with the number of modes in the array
+///
+/// \return An array of strings containing the names
+///
+////////////////////////////////////////////////////////////
+CSFML_AUDIO_API const char** sfSoundRecorder_getAvailableDevices(size_t* count);
+
+////////////////////////////////////////////////////////////
+/// \brief Get the name of the default audio capture device
+///
+/// This function returns the name of the default audio
+/// capture device. If none is available, NULL is returned.
+///
+/// \return The name of the default audio capture device (null terminated)
+///
+////////////////////////////////////////////////////////////
+CSFML_AUDIO_API const char* sfSoundRecorder_getDefaultDevice();
+
+////////////////////////////////////////////////////////////
+/// \brief Set the audio capture device
+///
+/// This function sets the audio capture device to the device
+/// with the given name. It can be called on the fly (i.e:
+/// while recording). If you do so while recording and
+/// opening the device fails, it stops the recording.
+///
+/// \param soundRecorder Sound recorder object
+/// \param The name of the audio capture device
+///
+/// \return sfTrue, if it was able to set the requested device
+///
+////////////////////////////////////////////////////////////
+CSFML_AUDIO_API sfBool sfSoundRecorder_setDevice(sfSoundRecorder* soundRecorder, const char* name);
+
+////////////////////////////////////////////////////////////
+/// \brief Get the name of the current audio capture device
+///
+/// \param soundRecorder Sound recorder object
+///
+/// \return The name of the current audio capture device
+///
+////////////////////////////////////////////////////////////
+CSFML_AUDIO_API const char* sfSoundRecorder_getDevice(sfSoundRecorder* soundRecorder);
 
 
 #endif // SFML_SOUNDRECORDER_H
--- a/src/SFML/Audio/SoundRecorder.cpp
+++ b/src/SFML/Audio/SoundRecorder.cpp
@@ -48,9 +48,9 @@ void sfSoundRecorder_destroy(sfSoundRecorder* soundRecorder)
 
 
 ////////////////////////////////////////////////////////////
-void sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate)
+sfBool sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate)
 {
-    CSFML_CALL(soundRecorder, start(sampleRate));
+    CSFML_CALL_RETURN(soundRecorder, start(sampleRate), sfFalse);
 }
 
 
@@ -71,7 +71,7 @@ unsigned int sfSoundRecorder_getSampleRate(const sfSoundRecorder* soundRecorder)
 ////////////////////////////////////////////////////////////
 sfBool sfSoundRecorder_isAvailable(void)
 {
-    return sf::SoundRecorder::isAvailable() ? sfTrue : sfFalse;
+    return sf::SoundRecorder::isAvailable();
 }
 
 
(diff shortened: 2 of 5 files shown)


include/SFML/Audio/SoundRecorder.h
```diff
@@ -132,5 +134,58 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
 ////////////////////////////////////////////////////////////
 CSFML_AUDIO_API void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval);
 
+////////////////////////////////////////////////////////////
+/// \brief Get a list of the names of all availabe audio capture devices
+///
+/// This function returns an array of strings, containing
+/// the names of all availabe audio capture devices.
```
LaurentGomila: You should state that the returned strings are null-terminated, this is important in C (this applies to the other functions as well).

include/SFML/Audio/SoundRecorder.h
```diff
@@ -132,5 +134,58 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
...
+///
+////////////////////////////////////////////////////////////
+CSFML_AUDIO_API const char** sfSoundRecorder_getAvailableDevices(size_t* count);
+
+////////////////////////////////////////////////////////////
+/// \brief Get the name of the default audio capture device
+///
+/// This function returns the name of the default audio
+/// capture device. If none is available, an empty string
+/// is returned.
```
LaurentGomila: Wouldn't NULL be more suitable for the default device? In C it is easier to test a NULL pointer than an empty string.

include/SFML/Audio/SoundRecorder.h
```diff
@@ -132,5 +134,58 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
...
+/// with the given \a name. It can be called on the fly (i.e:
+/// while recording). If you do so while recording and
+/// opening the device fails, it stops the recording.
+///
+/// \param soundRecorder Sound recorder object
+/// \param The name of the audio capture device
+///
+/// \return sfTrue, if it was able to set the requested device
+///
+/// \see getAvailableDevices, getDefaultDevice
```
LaurentGomila: Don't forget to adjust the function names here :)

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -80,3 +83,56 @@ void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTim
...
+////////////////////////////////////////////////////////////
+const char* sfSoundRecorder_getDefaultDevice()
+{
+    static std::string defaultDevice;
+
+    if (!defaultDevice.size())
+    {
+        defaultDevice = sf::SoundRecorder::getDefaultDevice();
+    }
+
```
LaurentGomila: Shorter notation:

```cpp
static std::string defaultDevice = sf::SoundRecorder::getDefaultDevice();
```

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -80,3 +83,56 @@ void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTim
...
+////////////////////////////////////////////////////////////
+const char** sfSoundRecorder_getAvailableDevices(size_t* count)
+{
+    static std::vector<std::string> stringDevices;
+    static std::vector<const char*> cstringDevices;
+
+    if (stringDevices.empty())
+    {
+        stringDevices = sf::SoundRecorder::getAvailableDevices();
+        cstringDevices.reserve(stringDevices.size());
```
LaurentGomila: This micro-optimization can probably be removed :p

include/SFML/Audio/SoundRecorder.h
```diff
@@ -132,5 +134,58 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
...
 
+////////////////////////////////////////////////////////////
+/// \brief Get a list of the names of all availabe audio capture devices
+///
+/// This function returns an array of strings, containing
+/// the names of all availabe audio capture devices.
+///
+/// \param count Pointer to a variable that will be filled with the number of modes in the array
+///
+/// \return An array of strings containing the names
```
LaurentGomila: You should add that in case of 0 device, it returns NULL (and not a valid pointer to an empty array).

LaurentGomila:
This is also how I would have done it. So apart from my minor comments, it looks good to me.

src/SFML/Audio/SoundRecorder.cpp
zsbzsb: If we did this then we could also shorten the device list function the same way. The only reason I didn't do this because I'm not sure when the initialization would happen. And the history of issues with global initialization order... :grin:

LaurentGomila: Static variables inside functions are initialized on the first function call. So it's perfectly safe.

But for the device lists, since initialization is not a one-liner, you can't apply the same shortcut.

zsbzsb: ```static std::vector<std::string> stringDevices = sf::SoundRecorder::getAvailableDevices();```

Could be changed to this^^ Also its good to know that they are definitely initialized on the first function call.

zsbzsb:
I pushed some more commits, feel free to review :smile:

src/SFML/Audio/SoundRecorder.cpp
LaurentGomila: Ah, yes this one can also be done directly. Just not the ```vector<const char*>```.

include/SFML/Audio/SoundRecorder.h
```diff
@@ -132,5 +134,56 @@ CSFML_AUDIO_API sfBool sfSoundRecorder_isAvailable(void);
 ////////////////////////////////////////////////////////////
 CSFML_AUDIO_API void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval);
 
+////////////////////////////////////////////////////////////
+/// \brief Get a list of the names of all availabe audio capture devices
+///
+/// This function returns an array of strings (null terminated),
+/// containing the names of all availabe audio capture devices.
+/// If no devices are available then NULL be returned.
```
LaurentGomila: "is" or "will be" :p

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -48,9 +48,12 @@ void sfSoundRecorder_destroy(sfSoundRecorder* soundRecorder)
...
 ////////////////////////////////////////////////////////////
-void sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate)
+sfBool sfSoundRecorder_start(sfSoundRecorder* soundRecorder, unsigned int sampleRate)
 {
-    CSFML_CALL(soundRecorder, start(sampleRate));
+    CSFML_CHECK(soundRecorder);
+    CSFML_CHECK(soundRecorder->This);
+
+    return soundRecorder->This.start(sampleRate);
 }
```
LaurentGomila: There's a CSFML_CALL_RETURN macro for this case.

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -80,3 +83,48 @@ void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTim
...
+
+    return !cstringDevices.empty() ? &cstringDevices[0] : NULL;
+}
+
+////////////////////////////////////////////////////////////
+const char* sfSoundRecorder_getDefaultDevice()
+{
+    static std::string defaultDevice = sf::SoundRecorder::getDefaultDevice();
+
+    return defaultDevice.size() ? defaultDevice.c_str() : NULL;
```
LaurentGomila: !empty() would be better than size().

src/SFML/Audio/SoundRecorderStruct.h
```diff
@@ -101,6 +101,7 @@ struct sfSoundRecorder
     }
 
     sfSoundRecorderImpl This;
+    std::string String;
```
LaurentGomila: I'd rather give this a more explicit name, such as ```DeviceName```.

zsbzsb:
All done :)

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -80,3 +80,48 @@ void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTim
...
 }
+
+
+////////////////////////////////////////////////////////////
+const char** sfSoundRecorder_getAvailableDevices(size_t* count)
+{
+    static std::vector<std::string> stringDevices = sf::SoundRecorder::getAvailableDevices();
+    static std::vector<const char*> cstringDevices;
+
+    if (cstringDevices.empty() && !stringDevices.empty())
```
zsbzsb: This is just a small optimization so we don't continually try to populate the cstringDevices vector when no devices are available.

LaurentGomila:
Looks good. Now it would be great if you could squash all these commits into a single clean one before I merge it.

zsbzsb:
Done

LaurentGomila:
:+1:

PR #59, opened 2015-02-24
Added support for joystick identification

zsbzsb:
What a big white box this is....

Lets fill it with https://github.com/LaurentGomila/SFML/commit/29c0f1491177d9125cc731f0c04bdf4c21c4650e

### DIFF
--- a/include/SFML/Window/Joystick.h
+++ b/include/SFML/Window/Joystick.h
@@ -29,6 +29,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Export.h>
+#include <SFML/Window/JoystickIdentification.h>
 
 
 ////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


LaurentGomila:
Shouldn't we store one std::string per joystick id, so that one call for joystick X doesn't invalidate the name of joystick Y?

zsbzsb:
I thought about that, but then I remembered we are in pre C++11 era. So there really isn't any hashmap we could use. Also I talked to @binary1248 and was told it is a generally accepted standard that any calls to a C function invalidates the previous returned pointers/values.

But if you can tell me a good way to store a specific string per id that doesn't involve copying the string when adding another id I'm all ears. I personally thought about using a vector but that may end up copying the strings and invalidating the pointer from c_str()... :(

http://en.cppreference.com/w/c/program/getenv
>This function is not required to be thread-safe. Another call to getenv, as well as a call to the POSIX functions setenv(), unsetenv(), and putenv() may invalidate the pointer returned by a previous call or modify the string obtained from a previous call. 
http://en.cppreference.com/w/c/chrono/ctime
> pointer to a static null-terminated character string holding the textual representation of date and time. The string may be shared between asctime and ctime, and may be overwritten on each invocation of any of those functions.

LaurentGomila:
Yeah I know, and I'm fine with that, I was just wondering if it would somehow be better to do it this way, since it is really one static string per joystick, and not a single shared buffer that will change on each call depending on the input parameters.

But let's keep it this way, it's ok.

zsbzsb:
> not a single shared buffer that will change on each call depending on the input parameters

Actually.. I looked through the code and it is possible for the joystick to change during execution.  At least the code is written as if the user unplugs/plugs in a different joystick the name would change.

LaurentGomila:
That's right.

PR #62, opened 2015-02-25
Added sensor API

zsbzsb:


include/SFML/Window/Sensor.h
```diff
@@ -0,0 +1,89 @@
...
+{
+    sfSensorTypeAccelerometer,    ///< Measures the raw acceleration (m/s�)
+    sfSensorTypeGyroscope,        ///< Measures the raw rotation rates (degrees/s)
+    sfSensorTypeMagnetometer,     ///< Measures the ambient magnetic field (micro-teslas)
+    sfSensorTypeGravity,          ///< Measures the direction and intensity of gravity, independent of device acceleration (m/s�)
+    sfSensorTypeUserAcceleration, ///< Measures the direction and intensity of device acceleration, independent of the gravity (m/s�)
+    sfSensorTypeOrientation,      ///< Measures the absolute 3D orientation (degrees)
+
+    sfSensorTypeCount             ///< Keep last -- the total number of sensor types
+} sfSensorType;
```
LaurentGomila: I was looking at how other enums are named in CSFML, and wondered if we could shorten these names somehow, either ```sfSensorAccelerometer``` or even ```sfAccelerometer```.

src/SFML/Window/Sensor.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    sf::Sensor::setEnabled(static_cast<sf::Sensor::Type>(sensor), enabled == sfTrue);
+}
+
+////////////////////////////////////////////////////////////
+sfVector3f sfSensor_getValue(sfSensorType sensor)
+{
+    sf::Vector3f value = sf::Sensor::getValue(static_cast<sf::Sensor::Type>(sensor));
+
+    return {value.x, value.y, value.z};
```
LaurentGomila: This is C++11, should be:

```cpp
sfVector3f value = {...};
return value;
```

LaurentGomila:
You forgot the new SensorChanged event.

src/SFML/Window/Sensor.cpp
eXpl0it3r: I guess, I'm missing something, but isn't this C++11 code as well?

LaurentGomila: No, this is brace initialization of an aggregate type (POD), this the same as when you initialize an array.

eXpl0it3r: Ah okay :smiley:

zsbzsb:
>I was looking at how other enums are named in CSFML, and wondered if we could shorten these names somehow

Done, now it is more inline with how the sfMouseButton enum is handled.

>This is C++11

Fixed, I thought kinda the opposite of @eXpl0it3r. Since I saw the brace initialization used for construction of a variable I thought I could return it directly and it wouldn't be C++11.

>You forgot the new SensorChanged event.

Could we do both events in a single new PR? I'm suggesting this because they both rely on ordering of the same enum and if stuff gets out of order the values will be wrong.

LaurentGomila:
> Could we do both events in a single new PR?

Of course, I don't care.

BlaXpirit:
https://github.com/LaurentGomila/SFML/issues/806

zsbzsb:
@BlaXpirit Ok I will make an update and change it to "^2".

PR #69, opened 2015-03-03
Fix pointer conversion warning when compiling without NDEBUG on Linux

jcowgill:
GCC doesn't like this NULL to integer conversion. Since this is C++, using 0 should work for both pointers and integers.

```
src/SFML/Window/Window.cpp:320:5: warning: converting to non-pointer type ‘sfWindowHandle {aka long unsigned int}’ from NULL [-Wconversion-null]
     CSFML_CHECK_RETURN(window, NULL);
     ^
```


zsbzsb:
Looks good to me, and its the way the render window implementation is done.

LaurentGomila:
Thanks for the help.

PR #73, opened 2015-03-22
Changed getSampleCount return type to sfUint64 to better match SFML

zsbzsb:
I noticed that when compiling I was getting an error of possible loss of data from a conversion. I checked it out and ```sf::SoundBuffer::getSampleCount()``` returns a Uint64, but the CSFML function returns a size_t which is only a uint32 on 32bit compilers.


Gyscos:
sfSoundBuffer_createFromSamples still takes a size_t as its sampleCount argument. Maybe we could use a sfUint64 for uniformity?

zsbzsb:
>Maybe we could use a sfUint64 for uniformity?

Yes, it should be changed too, I didn't think to check the ctors when changing the return type.

PR #74, opened 2015-03-30
Added support for context creation flags and texture/shader native handles

zsbzsb:
https://github.com/LaurentGomila/SFML/pull/779


LaurentGomila:
Why sfUint32 for native handles? The C++ API defines them as unsigned int.

zsbzsb:
Why do we even bother with sf::Uint32? The C++ API mixes them both interchangeably. Not to mention....

https://github.com/LaurentGomila/CSFML/blob/master/include/SFML/Config.h#L140
>typedef unsigned int sfUint32;

LaurentGomila:
> Why do we even bother with sf::Uint32 / sfUInt32?

Fixed size types are needed in many situations. The fact that all our target platforms map sf::Uint32 it to unsigned int is just an implementation detail, which may not be true anymore in the future.

What's wrong with simply using the same types?

zsbzsb:
>What's wrong with simply using the same types in CSFML?

Nothing at all. Commit updated.

PR #77, opened 2015-04-30
Use size_t instead of unsigned int to follow SFML change

Gyscos:
Count of objects and indices now use size_t in SFML. This pull request applies this change to CSFML.
It fixes compilation (ShapeImpl now correctly implements the pure virtual methods), and also correctly uses size_t in the other places.


zsbzsb:
It looks good to me.
:+1:

zsbzsb:
One last thing I would suggest is squashing your commits into a single commit. :wink:

Gyscos:
Squashed it is, sir.

PR #88, opened 2015-09-02
Added sfFtp_sendCommand

### LINKED ISSUE #87: Ftp::sendCommand is missing
[Ftp::sendCommand](http://www.sfml-dev.org/documentation/2.2/classsf_1_1Ftp.php#a44e095103ecbce175a33eaf0820440ff) was added in SFML 2.2, is missing from CSFML.

BlaXpirit:
Fixes #87 

It compiles but I haven't checked if it actually works.

### DIFF
--- a/include/SFML/Network/Ftp.h
+++ b/include/SFML/Network/Ftp.h
@@ -506,5 +506,25 @@ CSFML_NETWORK_API sfFtpResponse* sfFtp_download(sfFtp* ftp, const char* distantF
 ////////////////////////////////////////////////////////////
 CSFML_NETWORK_API sfFtpResponse* sfFtp_upload(sfFtp* ftp, const char* localFile, const char* destPath, sfFtpTransferMode mode);
 
+////////////////////////////////////////////////////////////
+/// \brief Send a command to the FTP server
+///
+/// While the most often used commands are provided as
+/// specific functions, this function can be used to send
+/// any FTP command to the server. If the command requires
+/// one or more parameters, they can be specified in
+/// \a parameter. Otherwise you should pass NULL.
+/// If the server returns information, you can extract it
+/// from the response using sfResponse_getMessage().
+///
+/// \param ftp       Ftp object
+/// \param command   Command to send
+/// \param parameter Command parameter
+///
+/// \return Server response to the request
+///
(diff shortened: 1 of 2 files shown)


include/SFML/Network/Ftp.h
```diff
@@ -506,5 +506,25 @@ CSFML_NETWORK_API sfFtpResponse* sfFtp_download(sfFtp* ftp, const char* distantF
 ////////////////////////////////////////////////////////////
 CSFML_NETWORK_API sfFtpResponse* sfFtp_upload(sfFtp* ftp, const char* localFile, const char* destPath, sfFtpTransferMode mode);
 
+////////////////////////////////////////////////////////////
+/// \brief Send a command to the FTP server
+///
+/// While the most often used commands are provided as member
+/// functions in the sfFtp class, this method can be used
+/// to send any FTP command to the server. If the command
```
zsbzsb: "provided as member functions in the sfFtp class" should be rewritten because we don't have member functions in C code. I would just write it as "provided through other functions".

include/SFML/Network/Ftp.h
```diff
@@ -506,5 +506,25 @@ CSFML_NETWORK_API sfFtpResponse* sfFtp_download(sfFtp* ftp, const char* distantF
...
+/// requires one or more parameters, they can be specified
+/// in \a parameter. Otherwise it should be an empty string.
```
zsbzsb: "Otherwise it should be an empty string" should probably read "Otherwise you should pass NULL".

zsbzsb:
Other than the few minor documentation things it looks good.

include/SFML/Network/Ftp.h
BlaXpirit: SFML has `""` as the default argument.

zsbzsb: Yes but in a C binding it makes more sense to pass NULL instead of an empty string. This is already the case in other parts of the code also. [See here](https://github.com/SFML/CSFML/blob/master/include/SFML/Graphics/Shader.h#L44) as an example.

zsbzsb:
Thanks! :+1:

PR #90, opened 2015-09-29
Added missing sfSoundBufferRecorder_XXX methods

### LINKED ISSUE #83: Missing methods on SoundBufferRecorder.
These functions on `SoundRecorder` are not available on `SoundBufferRecorder`:
```c
void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval);
sfBool sfSoundRecorder_setDevice(sfSoundRecorder* soundRecorder, const char* name);
const char* sfSoundRecorder_getDevice(sfSoundRecorder* soundRecorder);
```
It seems like at least `setDevice` and `getDevice` variants should be available.

zsbzsb:
Fixes #83. This PR only adds 2 new methods to set/get the audio device. I left out the processing interval as that is a protected method (only accessible when inheriting ```sf::SoundRecorder``` and since ```sfSoundBufferRecorder``` is already a specialized subclass it doesn't provide access to any of the other protected methods (callbacks).


zsbzsb:
I am going to merge this unless @LaurentGomila has anything to say.

LaurentGomila:
Looks good :+1:

Issue #54, opened 2015-01-06
Update documentation


eXpl0it3r:
I might do a PR later, but just so it doesn't get forgotten.

When generating the documentation there are quite a few errors generated, see [this Gist](https://gist.github.com/7e773d9ad1f4a7206d2e).

Also after SFML 2.2, we've [updated the doxygen file quite a bit](https://github.com/LaurentGomila/SFML/pull/750), maybe we should apply most of the changes from over there to CSFML as well.

zsbzsb:
Superseded by #92

PR #92, opened 2015-10-16
Corrected issues with documentation

zsbzsb:
As reported by @eXpl0it3r in #54. However I am not sure what exactly @eXpl0it3r did when he made changes to the actual doxygen files, so any chance you could add a commit to bugfix/doc_issues branch with the same changes?

### DIFF
--- a/include/SFML/Network/IpAddress.h
+++ b/include/SFML/Network/IpAddress.h
@@ -114,8 +114,7 @@ CSFML_NETWORK_API sfIpAddress sfIpAddress_fromInteger(sfUint32 address);
 /// from a host name.
 ///
 /// \param address Address object
-///
-/// \return String representation of the address
+/// \param string  String where the string representation will be stored
 ///
 ////////////////////////////////////////////////////////////
 CSFML_NETWORK_API void sfIpAddress_toString(sfIpAddress address, char* string);
--- a/include/SFML/Graphics/CircleShape.h
+++ b/include/SFML/Graphics/CircleShape.h
@@ -58,7 +58,7 @@ CSFML_GRAPHICS_API sfCircleShape* sfCircleShape_copy(const sfCircleShape* shape)
 ////////////////////////////////////////////////////////////
 /// \brief Destroy an existing circle Shape
 ///
-/// \param Shape Shape to delete
+/// \param shape Shape to delete
 ///
 ////////////////////////////////////////////////////////////
 CSFML_GRAPHICS_API void sfCircleShape_destroy(sfCircleShape* shape);
--- a/include/SFML/Graphics/ConvexShape.h
+++ b/include/SFML/Graphics/ConvexShape.h
@@ -58,7 +58,7 @@ CSFML_GRAPHICS_API sfConvexShape* sfConvexShape_copy(const sfConvexShape* shape)
 ////////////////////////////////////////////////////////////
 /// \brief Destroy an existing convex Shape
 ///
-/// \param Shape Shape to delete
+/// \param shape Shape to delete
 ///
 ////////////////////////////////////////////////////////////
 CSFML_GRAPHICS_API void sfConvexShape_destroy(sfConvexShape* shape);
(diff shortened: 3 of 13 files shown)


LaurentGomila:
You should separate parameter name changes from behaviour modifications. I have a lot to say about the latter :stuck_out_tongue:

zsbzsb:
Well then say them, because the code changes are being done to correct the docs. Can't fix one without fixing the other :wink:

include/SFML/Network/IpAddress.h
```diff
@@ -115,10 +115,10 @@ CSFML_NETWORK_API sfIpAddress sfIpAddress_fromInteger(sfUint32 address);
 ///
 /// \param address Address object
 ///
-/// \return String representation of the address
+/// \return String as a locale-dependant ANSI string
 ///
 ////////////////////////////////////////////////////////////
-CSFML_NETWORK_API void sfIpAddress_toString(sfIpAddress address, char* string);
+CSFML_NETWORK_API const char* sfIpAddress_toString(sfIpAddress address);
```
zsbzsb: But @LaurentGomila I know you are eye balling this. So go ahead and say it. :smile:

LaurentGomila:
How do the changes in the following functions relate to the documentation:
* sfIpAddress_toString
* sfTcpSocket_receive

??

Anyway, here are my remarks:
* you changed the behaviour of ```sfIpAddress_toString```, which now makes it non-reentrant and prevents callers from storing an reusing the result
* ```sfTcpSocket_receive``` will now crash if ```received``` is null

zsbzsb:
I guess I forgot to respond to this... :P

>How do the changes in the following functions relate to the documentation

Because to fix the documentation the functions need to be corrected.

```sfIpAddress_toString``` was changed so it is consistent with the rest of the library and the documentation. Regarding consistency, no other function in CSFML requires the user to allocate memory for the returned string. There is also the problem as it currently stands that the user has no idea how much memory should be allocated for the string. So unless we add another function so the user can know the string's length before calling ```sfIpAddress_toString``` changing it so it returns a pointer to our own buffer is the best choice.

>you changed the behaviour of ```sfIpAddress_toString```, which now makes it non-reentrant and prevents callers from storing and reusing the returned pointer

This is the same with all of the other CSFML functions that return a pointer to a string. None of them guarantee that the pointer will be valid after calling the function a second time. As far as I know it is standard practice to copy a string into your own buffer (that another function allocated) if you plan on keeping it around when calling the function again. And with this change the user will be able to know the exact length his buffer will need to be before allocating it (if he plans on copying the string to keep it around).

```sfTcpSocket_receive``` implementation was changed purely based on the fact it should never be called with ```received == NULL```. The function itself becomes pretty much becomes useless unless you know how much data was received and ```sf::TcpSocket::receive(...)``` function itself does not make that parameter optional.

> ```sfTcpSocket_receive``` will now crash if ```received``` is null

If any user wrote code that made the assumption that ```received``` could be null was relying on undefined behavior. Anywhere else in CSFML if an argument is optional the documentation explicitly states that NULL can be passed. See ```[sfShader_createFromStream](https://github.com/SFML/CSFML/blob/master/include/SFML/Graphics/Shader.h#L89)``` as one example.

LaurentGomila:
> sfTcpSocket_receive implementation was changed purely based on the fact it should never be called with received == NULL. The function itself becomes pretty much becomes useless unless you know how much data was received and sf::TcpSocket::receive(...) function itself does not make that parameter optional.

But... why change a possibly useless but totally safe behaviour, to a UB? Add to the doc that it can be null if you want.

zsbzsb:
I don't mind changing it back so received can be null as I was just trying to cleanup the code, but what about the toString method?

LaurentGomila:
> what about the toString method?

If you tell me that every other similar function works the same way (too lazy to check myself :stuck_out_tongue:), then it's fine.

zsbzsb:
Alright, I updated the PR, I removed the changes to the ```sfIpAddress_toString(...)``` as I wanted to merge this before tagging 2.3.x. I will save the API changes for when we are ready to tag 2.4.

@LaurentGomila  Would you mind doing a final review?

zsbzsb:
@LaurentGomila *bump* :wink:

LaurentGomila:
Sorry I'm really too busy. I trust you on this one :smiley_cat:

PR #93, opened 2015-11-13
Allow static linking to SFML libs on non Windows platforms

zsbzsb:
It was reported on SFML's IRC chat by @ShrewdSpirit that when configuring CSFML to link against static SFML libs (libsfml-xxx-s.a) FindSFML.cmake was not finding the libs. After some checking to make sure everything else was right I looked through CSFML's CMakeLists.txt files and noticed that `CSFML_LINK_SFML_STATICALLY` was ignored on non Windows platforms. After removing the check for the Windows platforms the issue disappeared and the scripts now properly find the static SFML libs.

This PR sets `CSFML_LINK_SFML_STATICALLY` to true by default on Windows and to false on other platforms. It also removes the platform check when setting `SFML_STATIC_LIBRARIES` and defining `SFML_STATIC`.


zsbzsb:
@LaurentGomila ? :wink:

LaurentGomila:
Does linking SFML statically to CSFML work fine on non-Windows platforms?

zsbzsb:
What are you trying to ask? Of course static linking works on linux (its why SFML allows it) - I'm not saying it is the recommended way to link stuff on non Windows platforms, but the CSFML cmake scripts should not force the user to use dynamic SFML libs on a non Windows platform. All that aside, having the scripts silently ignore ```CSFML_LINK_SFML_STATICALLY``` on a non Windows platform should definitely not be the default behavior.

LaurentGomila:
What I was trying to ask was simply "have you tested the stuff that you're about to enable?". You know, just in case.

zsbzsb:
I'm not sure what stuff you want to have tested, but this does allow you to link to SFML statically on non windows platforms (such as linux where the issue was discovered). If you don't have anything against it (maybe is there a better way to toggle the default setting per platform?) let's get this merged. :smile:

zsbzsb:
@LaurentGomila *bump* :wink:

LaurentGomila:
ok let's go :stuck_out_tongue:

PR #95, opened 2016-03-24
Fixed typo in readme

WillyGoat95:


LaurentGomila:
Thanks!

zsbzsb:
I must have read that readme god knows how many times and never caught it. :+1:

PR #101, opened 2016-08-07
Added return value to function sfSoundBufferRecorder_start

DJMaster:
With this patch sfSoundBufferRecorder_start return sfTrue if it was able to start recording.


LaurentGomila:
Thank you!

PR #115, opened 2016-10-26
Add sRgbCapable field to sfContextSettings, reduce code duplication

crumblingstatue:
Add a new `in` namespace where internal functions go, and add functions
there to factor out duplicated code for sfContextSettings.


eXpl0it3r:
Feels a bit odd to introduce a namespace in an otherwise C-like environment, but I'll let Laurent be the judge here.

For SFML we normally use `priv` as internal namespace. I think it might fit here as well.

crumblingstatue:
> Feels a bit odd to introduce a namespace in an otherwise C-like environment, but I'll let Laurent be the judge here.

The name is strictly for use by the internal implementation, which is implemented in C++.

> For SFML we normally use priv as internal namespace. I think it might fit here as well.

I can change that, if requested.

crumblingstatue:
The main rationale for using a namespace stems from using the same prefixes for the internal functions as in the public API. e.g. `sfRenderWindow_someInternalFunc` vs `sfRenderWindow_new`. While name conflicts are unlikely, I want to rule them out.

LaurentGomila:
Thanks for contributing 👍 

First, you should never mix two unrelated modifications in the same commit and PR. The tiny `sRgbCapable` thing could have been merged immediately, but now it will have to wait. So please create a separate PR for this, and if it depends on the other part, then wait until it is finalized first.

I don't think that this "code duplication" deserves so much attention. After all there's no logic, it's just initialization of a structure.
On top of that, a file named "Macros.hpp" is always wrong, whatever it contains :wink: It would better be named "ContextSettingsInternal.hpp" or whatever.

So... if it was me, I would leave things as they are.

crumblingstatue:
> The tiny sRgbCapable thing could have been merged immediately

It wouldn't have been so tiny, because I would have had to change a lot of copy-pasted code for initializing and modifying the sfContextSettings struct.

> I don't think that this "code duplication" deserves so much attention. After all there's no logic, it's just initialization of a structure.

Copy/paste related errors are common in software development. Right now, if the developer wants to add a new field to the sfContextSettings struct, he has to make sure to go through all code where the sfContextSettings struct is initialized/modified, and make sure to update each and every one properly.
This is a lot of tedious copy/paste work.

With the new code, the developer just has to change the 3 functions defined in `Internal/Window.h`.

> On top of that, a file named "Macros.hpp" is always wrong, whatever it contains :wink: It would better be named "ContextSettingsInternal.hpp" or whatever.

`Macros.hpp` is just a rename of `Internal.h`. `Internal.h` was not authored by me, and it just contains a bunch of macros for internal use. `Internal.h` isn't any more descriptive than `Internal/Macros.hpp`. I just renamed it, so I could put more internal code into an `Internal` folder, and `Internal/Macros.hpp` makes more sense than `Internal/Internal.h`. Again, `Macros.hpp` has nothing to do with  `ContextSettings`, so your `ContextSettingsInternal.hpp` suggestion for it is wrong.

LaurentGomila:
> Macros.hpp has nothing to do with ContextSettings, so your ContextSettingsInternal.hpp suggestion for it is wrong.

Doh! Sorry, it seems like I mixed things while reading the diff.

crumblingstatue:
Okay, let's clarify this. Are you not considering this for merging, unless I make 2 separate pull requests? Or did you just not have the time to review this yet?

LaurentGomila:
What I said previously, except about the name of the header, still applies.

crumblingstatue:
> What I said previously, except about the name of the header, still applies.

You didn't address any of my points in my reply, except the header.

LaurentGomila:
> You didn't address any of my points in my reply, except the header.

I think I did:

-
> please create a separate PR for this
> [...]
> I don't think that this "code duplication" deserves so much attention.
> [...]
> So... if it was me, I would leave things as they are.

-

> If that is the case, I refuse to add sRgbCapable as well, as I don't like being forced to write duplicate code without being able to refactor it.

Fair enough. I don't want to block improvements on CSFML, especially since I no longer work on it, and it's the community that keeps it alive. So let's go with your refactoring.

Here are a few comments about the implementation details though:
* as already said, in SFML we name the internal namespace "priv", not "in".
* any reason why these new functions are inlined?
* coding conventions not followed in several places, I let you find where ;)
* we don't have to follow the naming conventions of public functions in internal stuff, but if you feel like this is better anyway, then I'm ok with that
* I'd probably create a dedicated ContextSettings.h header now, instead of including Window.h everywhere you just need context settings.

crumblingstatue:
> I think I did:
> 
> >please create a separate PR for this
> >[...]
> >I don't think that this "code duplication" deserves so much attention.
> >[...]
> >So... if it was me, I would leave things as they are.

Those are your original comments on the PR. I replied to them, and you didn't address those replies.

> as already said, in SFML we name the internal namespace "priv", not "in"

Alright, I stated that if requested, I will change the namespace to `priv`, but it wasn't explicitly requested until now.

> any reason why these new functions are inlined?

I put their whole definition in a header, and (AFAIK) if they were included in multiple translation units, it would result in duplicate symbol errors when linking. But you are right, I could just put their definition in a `.cpp` file. The compiler is probably smart enough to inline it on its own.

> coding conventions not followed in several places, I let you find where 

Alright, I'll check

> we don't have to follow the naming conventions of public functions in internal stuff, but if you feel like this is better anyway, then I'm ok with that

You mean the `sfContextSettings_` prefix? I still want to encode in the name that this function is strongly related to `sfContextSettings`. Do you know a better way?

> I'd probably create a dedicated ContextSettings.h header now, instead of including Window.h everywhere you just need context settings.

Alright, I'll do that.

crumblingstatue:
Anyway, since you want me to do separate pull requests, I'll close this one.

Issue #118, opened 2016-10-31
Dynamic linking of csfml-graphics fails after #116


crumblingstatue:
#116 introduced functions that both `csfml-window` and `csfml-graphics` utilize.

It seems that when dynamically linking `csfml-graphics` to a project, it cannot find the functions added in #116. I assume this is due to how visibility in shared objects works, as this doesn't seem to affect building a static library of CSFML.

```
libcsfml-graphics.so: undefined reference to `priv::sfContextSettings_writeToCpp(sfContextSettings const&, sf::ContextSettings&)'
libcsfml-graphics.so: undefined reference to `priv::sfContextSettings_null()'
libcsfml-graphics.so: undefined reference to `priv::sfContextSettings_readFromCpp(sf::ContextSettings const&, sfContextSettings&)'
```

I know of two solutions:

1. Mark these 3 functions `CSFML_API_EXPORT`. This sets the visibility of the functions, so that `csfml-graphics` can see them. However, it seems to me that since the functions are in separate shared objects, inlining will never occur, which is probably less efficient than having them inlined. (Or maybe the compiler provides the uninlined code in the shared library, while also injecting inlined code where it deems appropriate? I don't know how this works, maybe someone can enlighten me.)

2. Mark these 3 functions inline, and provide their definition inline, as was the case originally in #115. This solves the linking problem, while allowing inlining to occur. I personally prefer this option. I heard people argue that forcing a function to be inline can make code less efficient, but I have two counter-points to this:
    1. Solution no. 1 disallows inlining entirely, which is probably worse than forcing inlining.
    2. The original code was also "inlined", copy-pasted code, so this should generate to the exact same code.

LaurentGomila:
We don't really care about efficiency in this specific case, so I'd just inline them. Exporting internal functions seems wrong.

crumblingstatue:
Alright, this is solved by #119

PR #119, opened 2016-10-31
Inline the functions in ContextSettingsInternal

crumblingstatue:
This solves a visibility problem when linking dynamically against
csfml-graphics.

Ref #118


DJMaster:
Just tested. Please go ahead and apply. Thanx.

LaurentGomila:
Thanks for the fix and for the feeback.

PR #133, opened 2019-01-09
Add NuGet package tools

Davipb:
This is a companion to [PR #160 in SFML.Net](https://github.com/SFML/SFML.Net/pull/160), as requested by @eXpl0it3r 

This PR adds all the machinery required to distribute CSFML as a NuGet package under `tools/nuget`.

On Windows, all of CSFML's dependencies are linked statically, generating a single `.dll` per module (plus OpenAL.dll). On Linux, however, this was not possible to do: setting `BUILD_SHARED_LIBS=0` on SFML and `CSFML_LINK_SFML_STATICALLY=1` on CSFML causes the compilation to fail, stating that an `.so` can't be linked with a `.a`. As such, CSFML, SFML, and all of SFML's dependencies need to be distributed as separate files.

This can be mitigated by setting the rpath of the `.so`s to point to the current directory, but that forces all native libraries to be on the same folder, and on the same package. Because of this, I was unable to split CSFML into one package per module like SFML.Net.

I don't have access to a Mac computer, so I was unable to create scripts for Mac. It should be possible to adapt `build.linux.sh` to work on Macs, although the rpath manipulation needs to be converted to [`@loader_path` or something similar](https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath)


MarkWilds:
Is it a big problem for the CSFML.Net package to just have all **.so** in the '**runtimes/<any_linux>/native/*.so**' up to including all SFML shared libs, excluding SFML dependencies?  
(users are expected to install the dependencies to run your SFML.net application on linux)

Windows would just have one big csfml.dll in the package located at '**runtimes/win-<x86,x64>/native/csfml.dll**'.

Davipb:
Well, that's kind of how it works already:

On linux, this package contains `libcsfml-(module).so` and `libsfml-(module).so` (with its versioned counterparts - `libsfml-(module).so.2.5`), and the user needs to have SFML's dependencies installed locally

On Windows, this package contains `csfml-(module).dll`, which is standalone (statically linked to SFML, SFML's dependencies, the MSVC C library, etc).

I can't put all of Windows' native libraries into a big `csfml.dll` because the C# code imports every module as a separate dll with `[DllImport("csfml-(module)")]`, and making some kind of `#if` condition to change it to just `csfml` requires separate C# libraries for Windows and Linux, which kind of defeats the point of .NET Standard

MarkWilds:
Oke awesome! Ship it!

tools/nuget/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+* `build.win.ps1`
+    * A [PowerShell](https://github.com/PowerShell/PowerShell) script to build Windows native libraries. By default, both `win-x86` and `win-x64` are built. To build only one specific Runtime Identifier, pass it to the script as a parameter named `RID` (`./build.win.ps1 -RID win-x64`)  
+    You must have Visual Studio 2017, cmake, and the Windows SDK 8.1 installed to run this script.
+
+* `build.linux.sh`
+    * A [Bash](https://www.gnu.org/software/bash/) script to build libraries for a generic Linux distribution. The [Runtime Identifier](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog) of the linux distribution must be specified as a parameter to the script.  
+    This generic script doesn't install any of its dependencies. As such, you must have cmake and make installed, as well as all [SFML Dependencies](https://www.sfml-dev.org/tutorials/2.5/compile-with-cmake.php#installing-dependencies) available to cmake.
+
+* `build.alpine-x64.sh`, `build.debian-x64.sh`, `build.fedora-x64.sh`
```
eXpl0it3r: Would it be possible to extend these scripts to automate the build-up and tear-down of the docker containers?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tools/nuget/SFML"]
+	path = tools/nuget/SFML
+	url = git@github.com:SFML/SFML.git
```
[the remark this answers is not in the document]
Davipb: The `.gitmodules` file doesn't show it, but the submodule is already pointing to the `2.5.x` tag in the SFML repository. You can check this by clicking at the SFML folder in [GitHub's interface](https://github.com/Davipb/CSFML/tree/nuget/tools/nuget) or by cloning the repo with `--recurse` and opening the `.git/modules/tools/nuget/SFML/HEAD` file: Both point to [commit 2f11710abc5aa478503a7ff3f9e654bd2078ebab](https://github.com/SFML/SFML/commit/2f11710abc5aa478503a7ff3f9e654bd2078ebab).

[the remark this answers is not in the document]
LaurentGomila: If we add SFML as a submodule, then shouldn't we put it at the root (or in a "extlibs" directory) rather than in tools/nuget? It would be part of the CSFML project, not only be used for the NuGet package.

tools/nuget/README.md
eXpl0it3r: What permission issues are there when binding the volume?

Binding the whole repo seems to make the most sense to me, as we don't really want to clone the repository again, plus at one point you need to bind something to the host anyway or use `scp` or similar to get the assemblies out of the docker image.

Right now, my main issue, that I don't exactly know how to start the various docker containers and what additional commands need to be executed before I can run the build scripts.

Davipb: On Linux, every file or directory created by the container will be owned by root, which can cause permission errors when deleting or moving them.

On Windows, host files/directories are mounted with fixed permissions and ownership, which can't be changed. This may cause issues if certain programs require specific permissions (the same way that ssh requires specific permissions for your keys). Also, the host permissions of files created inside the container are completely dependent on implementation details of Docker itself.

Bind-mounting the repo is also the solution I would go with, but it's good to keep possible permission issues in mind.

.gitmodules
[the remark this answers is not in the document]
Davipb: Well, right now it's only used by the NuGet package, since I didn't touch any of CSFML's files or cmake scripts, so that's why I put it in the `tools/nuget` folder.

I can move it to another folder if you decide that's better, the exact location of SFML doesn't matter to the NuGet build scripts.

tools/nuget/README.md
eXpl0it3r: I've never had permission issues with binding volumes, I've even bound locally mounted samba shares and it's working just fine. Then again my Docker host was always a Linux distro.

Wouldn't the idea here be to startup a new docker container, install what's needed, build SFML, build CSFML, copy assemblies and then shutdown/remove the container again? That way I don't see where a non-root user would want to copy/access anything or am I missing something here?

As for the Linux Docker > Windows host copy, not sure what kind of permission is added, but I'd say that regardless of the situation, the nuget building isn't something everyone on every possible system should be running, as such it's okay to expect that the user running it, has all the required rights etc.

.gitmodules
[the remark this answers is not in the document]
LaurentGomila: And would it be more work for you to *not* include SFML as a submodule? Or can we go without it, as @eXpl0it3r suggests? I think this kind of decision should be discussed in a more general context.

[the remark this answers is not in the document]
Davipb: It shouldn't be much trouble to not have SFML as a submodule, it would just require an extra command before the build scripts to clone SFML into a temporary folder.

The submodule was really my "first instinct" when thinking about a build that requires another git repository, but I understand how adding a submodule can be a big step for a repo. I'll update the build to clone SFML on-demand.

tools/nuget/README.md
Davipb: Yeah, I understand, we shouldn't have any issues in the case of the NuGet package. I'll expand the scripts to create the Docker containers by bind-mounting the repo.

eXpl0it3r: Thanks 🙂

Davipb:
Both suggestions have been implemented:
* SFML is no longer included as a submodule, and is instead cloned at `Build/SFML` during the build process (the repo isn't cloned if the folder is already there, to speed up repeated builds)
* The creation and removal of Docker containers has been automated with `build.docker.sh` and `build.docker.ps1`. The scripts that run inside the containers themselves have been renamed to `docker.(RID).sh`.

Davipb:
That's a vestige from when there was one package per module, thanks for catching it, it's fixed now 😃

PR #146, opened 2019-11-19
Added missing statement with sfShader_createFromX

Luhrel:
We can now call sfShader_createFromFile/Memory/Stream with only a
geometry file/memory/stream.

### DIFF
--- a/src/SFML/Graphics/Shader.cpp
+++ b/src/SFML/Graphics/Shader.cpp
@@ -60,8 +60,16 @@ sfShader* sfShader_createFromFile(const char* vertexShaderFilename, const char* 
         }
         else
         {
-            // vertex + geometry + fragment shaders
-            success = shader->This.loadFromFile(vertexShaderFilename, geometryShaderFilename, fragmentShaderFilename);
+            if (!vertexShaderFilename && !fragmentShaderFilename)
+            {
+                // geometry shader only
+                success = shader->This.loadFromFile(geometryShaderFilename, sf::Shader::Geometry);
+            }
+            else
+            {
+                // vertex + geometry + fragment shaders
+                success = shader->This.loadFromFile(vertexShaderFilename, geometryShaderFilename, fragmentShaderFilename);
+            }
         }
     }
 
@@ -102,8 +110,16 @@ sfShader* sfShader_createFromMemory(const char* vertexShader, const char* geomet
         }
         else
         {
-            // vertex + geometry + fragment shaders
-            success = shader->This.loadFromMemory(vertexShader, geometryShader, fragmentShader);
+            if (!vertexShader && !fragmentShader)
+            {
+                // geometry shader only
+                success = shader->This.loadFromMemory(geometryShader, sf::Shader::Geometry);
+            }
+            else
+            {
+                // vertex + geometry + fragment shaders
+                success = shader->This.loadFromMemory(vertexShader, geometryShader, fragmentShader);
+            }
         }
     }
 
@@ -148,11 +164,19 @@ sfShader* sfShader_createFromStream(sfInputStream* vertexShaderStream, sfInputSt
         }
         else
         {
-            // vertex + geometry + fragment shaders
-            CallbackStream vertexStream(vertexShaderStream);
             CallbackStream geometryStream(geometryShaderStream);
-            CallbackStream fragmentStream(fragmentShaderStream);
-            success = shader->This.loadFromStream(vertexStream, geometryStream, fragmentStream);
+            if (!vertexShaderStream && !fragmentShaderStream)
+            {
+                // geometry shader only
+                success = shader->This.loadFromStream(geometryStream, sf::Shader::Geometry);
+            }
+            else
+            {
+                // vertex + geometry + fragment shaders
+                CallbackStream vertexStream(vertexShaderStream);
+                CallbackStream fragmentStream(fragmentShaderStream);
+                success = shader->This.loadFromStream(vertexStream, geometryStream, fragmentStream);
+            }
         }
     }
 


eXpl0it3r:
I guess it's fair to add this given that it's the only interface for loading.

I'm a bit confused, why we have a different API for this, instead of doing something similar to what we have in SFML. @LaurentGomila do you remember?

In SFML when one of the shader fails to load, [all of them will be unloaded](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Shader.php#ab8c8b715b02aba2cf7c0a0e0c0984250).
But there we provide an `loadFromFile(const std::string& filename, Type type)` function, so you can load each type on it's own and define your own failure strategy.

LaurentGomila:
We can have a single loading function in CSFML and SFML.Net, because we can pass null for shaders we don't want to load. This is not possible in C++ with arguments taken by const reference.

The C++ function that takes a single source and the type of shader, is not made for loading all shader components separately, but to create a single component shader -- the two others are reset.

Regarding this PR, I think it's necessary. It seems like when geometry shaders were added, the patch that was made to support them in C/.Net APIs was not well written and did not support all cases.

Luhrel:
Something like this: [dd15e59](https://github.com/SFML/CSFML/pull/146/commits/dd15e5960878ec512cda36ef307a9a8063266993) ?

LaurentGomila:
> Something like this: dd15e59 ?

What I meant is that we *don'*t need this overload in C and .Net 😉

Luhrel:
> > Something like this: [dd15e59](https://github.com/SFML/CSFML/commit/dd15e5960878ec512cda36ef307a9a8063266993) ?
> 
> What I meant is that we _don't_ need this overload in C and .Net wink

Oopsy doopsy.

Now fixed.

src/SFML/Graphics/Shader.cpp
```diff
@@ -60,8 +60,16 @@ sfShader* sfShader_createFromFile(const char* vertexShaderFilename, const char*
...
         else
         {
-            // vertex + geometry + fragment shaders
-            success = shader->This.loadFromFile(vertexShaderFilename, geometryShaderFilename, fragmentShaderFilename);
+            if (!vertexShaderFilename && !fragmentShaderFilename)
+            {
+                // geometry shader only
+                success = shader->This.loadFromFile(geometryShaderFilename, sf::Shader::Geometry);
+            }
+            else
```
eXpl0it3r: What about the situation where either vertexShaderFilename or fragmentShaderFilename is not set, i.e. shouldn't there be three cases as above?

src/SFML/Graphics/Shader.cpp
```diff
@@ -102,8 +110,16 @@ sfShader* sfShader_createFromMemory(const char* vertexShader, const char* geomet
...
         else
         {
-            // vertex + geometry + fragment shaders
-            success = shader->This.loadFromMemory(vertexShader, geometryShader, fragmentShader);
+            if (!vertexShader && !fragmentShader)
+            {
+                // geometry shader only
+                success = shader->This.loadFromMemory(geometryShader, sf::Shader::Geometry);
+            }
+            else
```
eXpl0it3r: What about the situation where either vertexShaderFilename or fragmentShaderFilename is not set, i.e. shouldn't there be three cases as above?

src/SFML/Graphics/Shader.cpp
```diff
@@ -148,11 +164,19 @@ sfShader* sfShader_createFromStream(sfInputStream* vertexShaderStream, sfInputSt
...
-            CallbackStream vertexStream(vertexShaderStream);
             CallbackStream geometryStream(geometryShaderStream);
-            CallbackStream fragmentStream(fragmentShaderStream);
-            success = shader->This.loadFromStream(vertexStream, geometryStream, fragmentStream);
+            if (!vertexShaderStream && !fragmentShaderStream)
+            {
+                // geometry shader only
+                success = shader->This.loadFromStream(geometryStream, sf::Shader::Geometry);
+            }
+            else
```
eXpl0it3r: What about the situation where either vertexShaderFilename or fragmentShaderFilename is not set, i.e. shouldn't there be three cases as above?

src/SFML/Graphics/Shader.cpp
Luhrel: I didn't implemented that because SFML doesn't propose the case with geometryShaderFilename + [vertexShaderFilename/fragmentShaderFilename].
To load every case, we should have 7 possibilities.

LaurentGomila: Yes, the C++ code does not support all combinations. To be honest, I think we've reached the limits of the current design of Shader::loadFromXxx functions; if it was written now, I would do it differently.

eXpl0it3r: Not sure I fully follow. It's not possible with a single loadFromFile call, but if you want, you can call `loadFromFile(filename, type)` and get all com combinations. So CSFML should allow all combinations as well, no?

Luhrel: See [
3f3d623 ](https://github.com/SFML/CSFML/pull/146/commits/3f3d623872b6085cdd3cc25cc9c5e1e06f379b0d) ?

LaurentGomila: > It's not possible with a single loadFromFile call, but if you want, you can call loadFromFile(filename, type) and get all combinations

No no no, this overload loads a single component *and resets the two others*. When geometry shaders didn't exist, it was used to create a program with either a vertex or a fragment shader *only*.

eXpl0it3r: Ah okay, I see now. So this still leaves the questions:

- Should CSFML support all the cases?
- Or should CSFML only support the cases SFML supports?
- Should SFML be updated to also support all cases?

LaurentGomila: > Should CSFML support all the cases?

How could it support cases that are not implemented in the C++ API? 😁 

> Should SFML be updated to also support all cases?

That sounds like he wisest solution. Unless some combinations don't make sense?

eXpl0it3r:
As this commit is included in the other PR #148 I'll close this one. The other PR should be merged soon 🙂

PR #150, opened 2020-01-11
Prevent AnyCPU builds on .NET Framework

### LINKED ISSUE #149: CSFML NuGet Packages
As per https://github.com/SFML/SFML.Net/issues/174, CSFML doesn't currently provide "plug-and-play" functionality for .NET Framework (.NET Standard 1.X) style libs. This would need to be resolved in the CSFML NuGet package, as SFML.NET relies upon runtimes supplied by CSFML.

Additionally, we need to look at ways to provide CI and automated building of the CSFML NuGet package to cover more platforms (Mac OS is the glaring omission) as well as avoid having to manually package CSFML updates.

Davipb:
As discussed in SFML/SFML.Net#174, the Nuget CSFML  package doesn't copy its native files correctly when building with AnyCPU on the .NET Framework.

Given that the .NET Framework is being deprecated soon, this PR adds an MSBuild file to the package that detects when AnyCPU is being used and instructs the user to either switch to .NET Core or change platforms to x86/x64, failing the build.

A property is provided to override this error in case the user wants to copy and load CSFML's files manually at runtime.

Additionally, the deprecated `PackageIconUrl` property was replaced with `PackageIcon` pointing to a local PNG.

This should fix #149 and SFML/SFML.Net#174


eXpl0it3r (approved):

Davipb:
The `.props` files are inside of the `net` folder, which *used* to be the [Target Framework Moniker](https://docs.microsoft.com/en-us/nuget/reference/target-frameworks#supported-frameworks) for the .NET Framework (while `netcoreapp` was the TFM for .NET Core), meaning that they would only be loaded while building for the .NET Framework.

But now with the rebranding of .NET Core to just .NET going forward, it seems like they decided to also rename the TFM for .NET Core to just `net`, meaning it now refers to both the .NET Framework and .NET 5+. This is a really baffling move that I just discovered while researching the cause of this issue and is probably going to break a lot of stuff. 

It seems like you can target just the .NET Framework now by using the `netframework` TFM, but I need to do some experimenting to confirm this. I'll reply back when I know more.

eXpl0it3r:
Thanks for looking into it. Do you want to send a PR or should I write it?

Davipb:
Fastests PR in the west 🤠

PR #158, opened 2021-04-14
Update CSFML version to 2.5.1

eXpl0it3r:
- Add macOS build script
- Update existing nuget build script

## ToDo

- [x] Fix AnyCPU warning for .NET 5
- [x] Fix framework warning with new NuGet package (fixed with #159)
- [x] ~Get SFML.Net working on macOS~ will be done as part of #161
- [ ] Tag new release
- [ ] Publish new NuGet package
- [ ] Update SFML website


eXpl0it3r:
Test NuGet package: [CSFML.2.5.1.nupkg.zip](https://github.com/SFML/CSFML/files/6313942/CSFML.2.5.1.nupkg.zip)

eXpl0it3r:
@Davipb maybe you have some idea on what I need to do, so that CSFML is found on macOS. It just seems to look for *.DLLs instead of the provided *.dylib 🤔

Davipb:
Analyzing the compiled binaries from the test NuGet package you provided, it seems that the libraries are trying to load versioned names of its dependencies (e.g. `libsfml-system.2.5.dylib`) but the NuGet only contains unversioned names (e.g. `libsfml-system.dylib`):

```
Load command #11 (size:  56, offset: 1560): load_dylib
  name: @rpath/libsfml-audio.2.5.dylib
            time stamp: 0x00000002
       current version: 0x00020501
  comptibility version: 0x00020500

Load command #12 (size:  56, offset: 1616): load_dylib
  name: @rpath/libsfml-system.2.5.dylib
            time stamp: 0x00000002
       current version: 0x00020501
  comptibility version: 0x00020500
```

I had the same issue when building for Linux and "solved" it by including multiple copies of each library in the package, versioned and unversioned (NuGet doesn't support symlinks sadly). Does it work if you do the same for macOS?

eXpl0it3r:
Yeah, I was thinking about this as well, but wasn't sure since it was complaining about missing dylib but dlls. I'll rebuild and try again.

eXpl0it3r:
I think that worked. The last challenge is getting the RPath right. The trick with LDFLAGS as on Linux somehow didn't work, meaning the same command just fails on macOS. I'll try to play around with it more, but if you have some idea, let me know.

Davipb:
It seems that the macOS equivalent of `$ORIGIN` is either `@executable_path` or `@loader_path`. Maybe try replacing `-D 'CMAKE_INSTALL_RPATH=$ORIGIN'` in lines 120 and 148 of the build script with either `-D 'CMAKE_INSTALL_RPATH=@executable_path'` or  `-D 'CMAKE_INSTALL_RPATH=@loader_path'`?

The CMAKE flag itself is definitely working, since `$ORIGIN` was embedded as an rpath into the libraries:
```
Load command #15 (size:  24, offset: 1776): rpath
    $ORIGIN
```

Davipb:
You might be able to get some extra info with probe tracing by setting the `COREHOST_TRACE` environment variable to 1 (and `COREHOST_TRACE_VERBOSITY` to something between 1 and 4 to adjust the verbosity level). It prints a lot of stuff, but what you're looking for are these:
<details>
<summary>COREHOST_TRACE output fragment</summary>

```
Reconciling library CSFML/2.5.0
Parsed native deps entry 0 for asset name: csfml-Audio from package: CSFML, library version: 2.5.0, relpath: runtimes/win-x64/native/csfml-Audio.dll, assemblyVersion , fileVersion 0.0.0.0
Parsed native deps entry 1 for asset name: csfml-Graphics from package: CSFML, library version: 2.5.0, relpath: runtimes/win-x64/native/csfml-Graphics.dll, assemblyVersion , fileVersion 0.0.0.0
Parsed native deps entry 2 for asset name: csfml-System from package: CSFML, library version: 2.5.0, relpath: runtimes/win-x64/native/csfml-System.dll, assemblyVersion , fileVersion 0.0.0.0
Parsed native deps entry 3 for asset name: csfml-Window from package: CSFML, library version: 2.5.0, relpath: runtimes/win-x64/native/csfml-Window.dll, assemblyVersion , fileVersion 0.0.0.0
Parsed native deps entry 4 for asset name: openal32 from package: CSFML, library version: 2.5.0, relpath: runtimes/win-x64/native/openal32.dll, assemblyVersion , fileVersion 0.0.0.0
```
</details>

Alternatively, you could make a fake NuGet package with different code for each RID and run it to check what RID the .NET Core host is using. Here's one I made: [lib.1.0.0.nupkg.zip](https://github.com/SFML/CSFML/files/6465910/lib.1.0.0.nupkg.zip), and here's how you use it (it should print the detected RID or 'unknown' if neither was used):
<details>
<summary>lib.1.0.0 usage</summary>

```csharp
using System;
using lib;

namespace netcoreproj
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(RidDetector.RID);
        }
    }
}
```
</details>

Also checking online, some people claim that `@executable_path` and `@loader_path` aren't exactly equivalent to `$ORIGIN` and check paths relative to the main executable (i.e. `dotnet`) instead of the `.dylib`. Does it work if you put all `.dylib` files in the working directory before launching the program? If so, then the .NET runtime is loading CSFML just fine, but the OS is unable to find CSFML's dependencies (e.g. SFML), and fails the load. If that's the case, we might have to rethink how we do things in macOS.

Sadly I don't have access to a macOS machine, so I don't think I'll be able to help with more practical debugging.

eXpl0it3r:
I can't figure the macOS builds out at this point.

For now, I'll just release CSFML 2.5.1 without macOS builds and then maybe I'll find some more time to iron this out further for CSFML 2.5.2 or CSFML 2.6...

PR #170, opened 2022-02-14
changed min cmake ver and CMAKE_SOURCE_DIR to PROJECT_SOURCE_DIR

brightprogrammer:
There was a build error when building CSFML stating that it failed to find files in `cmake` folder. Also there were some warnings for cmake version so changed cmake version too. This fixed the build for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.0.2)
 
 # define a macro that helps defining an option
 macro(csfml_set_option var default type docstring)
@@ -17,7 +16,7 @@ csfml_set_option(CMAKE_BUILD_TYPE Release STRING "Choose the type of build (Debu
 project(CSFML)
 
 # include the configuration file
-include(${CMAKE_SOURCE_DIR}/cmake/Config.cmake)
+include(${PROJECT_SOURCE_DIR}/cmake/Config.cmake)
 
 # setup version numbers
 set(VERSION_MAJOR 2)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 
-cmake_minimum_required(VERSION 2.8.3)
+# version 2.8.3 is deprecated so switch to 3.0
```
eXpl0it3r, line 2: Please remove this comment

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
...
+cmake_minimum_required(VERSION 3.0)
```
eXpl0it3r, line 3: Can you make it `3.0.2` so it's inline with SFML 2.5.1

eXpl0it3r (changes requested):
Thanks for the change!

Just two small comments 😉

CMakeLists.txt
brightprogrammer: Sure, will do that

CMakeLists.txt
brightprogrammer: ok it's done

eXpl0it3r:
Can you change the following line

https://github.com/SFML/CSFML/blob/317f58fa318450265b18657ce94c7224cae4c870/cmake/Config.cmake#L38

to something like this?

```
if(CMAKE_CXX_COMPILER MATCHES "clang[+][+]" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```

Not sure if it's the newer CMake version, but seems like it fails to detect AppleClang on macOS now.

----

Additionally, could you squash the commits into one?

brightprogrammer:
I'm sorry, I'm not that experienced with git, I think the squashing didn't go well but I made the requested changes

brightprogrammer:
> I'm sorry, I'm not that experienced with git, I think the squashing didn't go well but I made the requested changes

ok squashing worked, can you please check now? :smile:

brightprogrammer:
is it possibble that the problem is in the machine on which builds are running on?

eXpl0it3r (approved):

eXpl0it3r:
I managed to get it working, by using the code from SFML's 2.6.x branch 😄

eXpl0it3r:
Thanks a lot for this fix! 🙂

Issue #178, opened 2022-12-30
Consider adding a Default RenderState


eXpl0it3r:
It's currently unclear what a default RenderState would look like, so it might make sense to have a default RenderState that can be used as basis and then modified as need be.

See discussion thread [here](https://en.sfml-dev.org/forums/index.php?topic=9767.msg66807#msg66807) (and [here](https://en.sfml-dev.org/forums/index.php?topic=28827.0)).

Bqleine:
> See discussion thread [here](https://en.sfml-dev.org/forums/index.php?topic=9767.msg66807#msg66807) (and [here](https://en.sfml-dev.org/forums/index.php?topic=28827.0)).

Oh hey that's me! I'm gonna try to implement it tonight.

PR #180, opened 2023-01-05
Add function for default RenderState (#178)

Bqleine:
Took me a little while to get the environment working but I made it!
I tried to do my best to follow how everything else was made, I think this is the right way ?

Either way tell me if anything is wrong and have a nice day :)


Bqleine:
Can't have designated initializers in detroit 😔

eXpl0it3r (approved):

eXpl0it3r:
Thank you for adding this! 🙂

PR #189, opened 2023-06-27
Require SFML 2.6

tintou:
It is technically already the case since commits e606b52c39516d0c29e62bad1dd851a929b9ab95 and 7d2ebaf265140ee6d2ff0eb469ec36118d3cccc8


eXpl0it3r (approved):
Ah, I forgot to check if we request a version somewhere else when merging #189 

Thanks for this fix!

PR #190, opened 2023-07-03
Implement `sf::Rect<T>::getPosition()` and `sf::Rect<T>::getSize()`

ChrisThrasher:
Implements https://github.com/SFML/SFML/pull/1047

### DIFF
--- a/src/SFML/Graphics/Rect.cpp
+++ b/src/SFML/Graphics/Rect.cpp
@@ -30,6 +30,21 @@
 #include <SFML/Internal.h>
 
 
+namespace
+{
+sfVector2f toCType(const sf::Vector2f& vector)
+{
+    sfVector2f vec = {vector.x, vector.y};
+    return vec;
+}
+
+sfVector2i toCType(const sf::Vector2i& vector)
+{
+    sfVector2i vec = {vector.x, vector.y};
+    return vec;
+}
+}
+
 ////////////////////////////////////////////////////////////
 /// Check if a point is inside a rectangle's area
 ////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
@dogunbound mind reviewing this?

src/SFML/Graphics/Rect.cpp
```diff
@@ -30,6 +30,21 @@
 #include <SFML/Internal.h>
 
 
+namespace
+{
+sfVector2f toCType(const sf::Vector2f& vector)
+{
+    sfVector2f vec = {vector.x, vector.y};
+    return vec;
```
dogunbound, line 38: Why use temporary variables here? Why not:
```c++
return {vector.x, vector.y};
```

src/SFML/Graphics/Rect.cpp
```diff
@@ -30,6 +30,21 @@
...
+}
+
+sfVector2i toCType(const sf::Vector2i& vector)
+{
+    sfVector2i vec = {vector.x, vector.y};
+    return vec;
```
dogunbound, line 44: Why use temporary variables here? Why not:
```c++
return {vector.x, vector.y};
```

src/SFML/Graphics/Rect.cpp
ChrisThrasher: That requires C++11. Because SFML 2 uses C++03 I'm assuming that CSFML 2 inherits that same restriction.

dogunbound: Ah I see. Then i'll approve of this.

eXpl0it3r (approved):

PR #191, opened 2023-07-03
Implement min and max blend modes

ChrisThrasher:
Implements https://github.com/SFML/SFML/pull/1756

### DIFF
--- a/src/SFML/Graphics/BlendMode.cpp
+++ b/src/SFML/Graphics/BlendMode.cpp
@@ -41,5 +41,13 @@ const sfBlendMode sfBlendMultiply = { sfBlendFactorDstColor, sfBlendFactorZero, 
                                       sfBlendFactorDstColor, sfBlendFactorZero, sfBlendEquationAdd };
 
 ////////////////////////////////////////////////////////////
+const sfBlendMode sfBlendMin = { sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMin,
+                                 sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMin };
+
+////////////////////////////////////////////////////////////
+const sfBlendMode sfBlendMax = { sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMax,
+                                 sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMax };
+
+////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/BlendMode.cpp
```diff
@@ -40,6 +40,14 @@ const sfBlendMode sfBlendAdd = { sfBlendFactorSrcAlpha, sfBlendFactorOne, sfBlen
 const sfBlendMode sfBlendMultiply = { sfBlendFactorDstColor, sfBlendFactorZero, sfBlendEquationAdd,
                                       sfBlendFactorDstColor, sfBlendFactorZero, sfBlendEquationAdd };
 
+////////////////////////////////////////////////////////////
+const sfBlendMode sfBlendMin = { sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMin,
+                                 sfBlendFactorDstColor, sfBlendFactorZero, sfBlendEquationAdd };
+
+////////////////////////////////////////////////////////////
+const sfBlendMode sfBlendMmax = { sfBlendFactorOne, sfBlendFactorOne, sfBlendEquationMax,
+                                  sfBlendFactorDstColor, sfBlendFactorZero, sfBlendEquationAdd };
```
eXpl0it3r, line 49: Are you sure this is correct?

The C++ code is:
```cpp
const BlendMode BlendMin(BlendMode::One, BlendMode::One, BlendMode::Min);
const BlendMode BlendMax(BlendMode::One, BlendMode::One, BlendMode::Max);
```

And the overloaded constructor is implemented as:
```cpp
lendMode::BlendMode(Factor sourceFactor, Factor destinationFactor, Equation blendEquation) :
colorSrcFactor(sourceFactor),
colorDstFactor(destinationFactor),
colorEquation (blendEquation),
alphaSrcFactor(sourceFactor),
alphaDstFactor(destinationFactor),
alphaEquation (blendEquation)
{

}
```

Which would mean, for a three-parameter constructor, we'd use the same values for color and well as alpha properties, but your code uses different values. 🤔

eXpl0it3r (changes requested):

src/SFML/Graphics/BlendMode.cpp
ChrisThrasher: Yeah that seems right. I'll push a fix

eXpl0it3r (approved):

PR #192, opened 2023-07-04
Implement fetching bold kerning

dogunbound:
C doesn't have function overloading, so I had to resort to function name signature changes to resemble overloading. I think this is an ok way to do this

### DIFF
--- a/src/SFML/Graphics/Font.cpp
+++ b/src/SFML/Graphics/Font.cpp
@@ -122,6 +122,13 @@ float sfFont_getKerning(const sfFont* font, sfUint32 first, sfUint32 second, uns
 
 
 ////////////////////////////////////////////////////////////
+float sfFont_getBoldKerning(const sfFont* font, sfUint32 first, sfUint32 second, unsigned int characterSize)
+{
+    CSFML_CALL_RETURN(font, getKerning(first, second, characterSize, true), 0);
+}
+
+
+////////////////////////////////////////////////////////////
 float sfFont_getLineSpacing(const sfFont* font, unsigned int characterSize)
 {
     CSFML_CALL_RETURN(font, getLineSpacing(characterSize), 0);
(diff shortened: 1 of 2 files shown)


src/SFML/Graphics/Font.cpp
```diff
@@ -121,6 +121,13 @@ float sfFont_getKerning(const sfFont* font, sfUint32 first, sfUint32 second, uns
 }
 
 
+////////////////////////////////////////////////////////////
+float sfFont_getBoldKerning(const sfFont* font, sfUint32 first, sfUint32 second, unsigned int characterSize)
+{
+    CSFML_CALL_RETURN(font, getKerning(first, second, characterSize, sfFalse == sfTrue), 0);
```
eXpl0it3r, line 127: The comparison with `sfTrue` makes sense if you have a `bool` value, here it just seems confusing, plus it's actually wrong, because we need to pass in `true` and `sfFalse == sfTrue` always evaluates to `false.

```suggestion
    CSFML_CALL_RETURN(font, getKerning(first, second, characterSize, true), 0);
```

eXpl0it3r (changes requested):

src/SFML/Graphics/Font.cpp
dogunbound: c doesn't have true and false boolean values. I did it with 1. 

I just realized that today. I mixed that up in my head when I was writing it.

build.sh
ChrisThrasher, line 1: Could you please remove this file?

ChrisThrasher (changes requested):

ChrisThrasher (approved):

dogunbound:
Finally got it right. lol. I'll have to be a bit more thorough before I put up a PR.

src/SFML/Graphics/Font.cpp
```diff
@@ -121,6 +121,13 @@ float sfFont_getKerning(const sfFont* font, sfUint32 first, sfUint32 second, uns
 }
 
 
+////////////////////////////////////////////////////////////
+float sfFont_getBoldKerning(const sfFont* font, sfUint32 first, sfUint32 second, unsigned int characterSize)
+{
+    CSFML_CALL_RETURN(font, getKerning(first, second, characterSize, 1), 0);
```
eXpl0it3r, line 127: We are writing C++ code here, so you can easily use `true` instead of `1`

eXpl0it3r (approved):

PR #195, opened 2023-07-04
Implement `sf::Image::saveToMemory`

ChrisThrasher:
Implements https://github.com/SFML/SFML/pull/1669

This one got complicated since we have to take a `std::vector` and convert it into something that can be returned from a C function. I hope I did it well. It does incur the cost of copying the entire buffer which is unfortunate but there is no way to tell a vector to relinquish control of its contents so copying the vector into a `malloc`'d buffer seems like my only option.

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -30,6 +30,9 @@
 #include <SFML/Internal.h>
 #include <SFML/CallbackStream.h>
 
+#include <cstdlib>
+#include <cstring>
+
 ////////////////////////////////////////////////////////////
 sfImage* sfImage_create(unsigned int width, unsigned int height)
 {
@@ -128,6 +131,28 @@ void sfImage_destroy(sfImage* image)
 sfBool sfImage_saveToFile(const sfImage* image, const char* filename)
 {
     CSFML_CALL_RETURN(image, saveToFile(filename), sfFalse);
+}
+
+
+////////////////////////////////////////////////////////////
+sfBool sfImage_saveToMemory(const sfImage* image, unsigned char** buffer, size_t* size, const char* format)
+{
+    CSFML_CHECK_RETURN(image, sfFalse);
+
+    std::vector<sf::Uint8> output;
+    if (image->This.saveToMemory(output, format)){
+        // Create new buffer then return address and size to user
+        *buffer = reinterpret_cast<unsigned char*>(std::malloc(output.size()));
+        *size = output.size();
+
+        // Copy vector contents into buffer
+        std::memcpy(*buffer, &output[0], output.size());
+        return sfTrue;
+    }
+
+    *buffer = NULL;
+    *size = 0;
+    return sfFalse;
 }
 
 
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -139,14 +139,10 @@ sfBool sfImage_saveToMemory(const sfImage* image, unsigned char** buffer, size_t
 {
     CSFML_CHECK_RETURN(image, sfFalse);
 
-    std::vector<sf::Uint8> output;
+    static std::vector<sf::Uint8> output;
     if (image->This.saveToMemory(output, format)){
-        // Create new buffer then return address and size to user
-        *buffer = reinterpret_cast<unsigned char*>(std::malloc(output.size()));
+        *buffer = &output[0];
         *size = output.size();
-
-        // Copy vector contents into buffer
-        std::memcpy(*buffer, &output[0], output.size());
         return sfTrue;
     }
 
--- a/include/SFML/Graphics/Image.h
+++ b/include/SFML/Graphics/Image.h
@@ -156,8 +156,33 @@ CSFML_GRAPHICS_API void sfImage_destroy(sfImage* image);
 ///
 /// \return sfTrue if saving was successful
 ///
+/// \see sfImage_saveToMemory
+///
 ////////////////////////////////////////////////////////////
 CSFML_GRAPHICS_API sfBool sfImage_saveToFile(const sfImage* image, const char* filename);
+
+////////////////////////////////////////////////////////////
+/// \brief Save the image to a buffer in memory
+///
+/// The format of the image must be specified.
+/// The supported image formats are bmp, png, tga and jpg.
+/// This function fails if the image is empty, or if
+/// the format was invalid.
+///
+/// The array pointed to by \p buffer must be freed to prevent
+/// a memory leak.
+///
+/// \param image  Image object
+/// \param buffer Address of buffer to fill with encoded data
+/// \param size   Size of buffer
+/// \param format Encoding format to use
+///
+/// \return True if saving was successful
+///
+/// \see sfImage_saveToFile
+///
+////////////////////////////////////////////////////////////
+CSFML_GRAPHICS_API sfBool sfImage_saveToMemory(const sfImage* image, unsigned char** buffer, size_t* size, const char* format);
 
 ////////////////////////////////////////////////////////////
 /// \brief Return the size of an image
--- a/include/SFML/Graphics/Image.h
+++ b/include/SFML/Graphics/Image.h
@@ -169,9 +169,6 @@ CSFML_GRAPHICS_API sfBool sfImage_saveToFile(const sfImage* image, const char* f
 /// This function fails if the image is empty, or if
 /// the format was invalid.
 ///
-/// The array pointed to by \p buffer must be freed to prevent
-/// a memory leak.
-///
 /// \param image  Image object
 /// \param buffer Address of buffer to fill with encoded data
 /// \param size   Size of buffer


src/SFML/Graphics/Image.cpp
```diff
@@ -131,6 +134,27 @@ sfBool sfImage_saveToFile(const sfImage* image, const char* filename)
...
 
+////////////////////////////////////////////////////////////
+sfBool sfImage_saveToMemory(const sfImage* image, unsigned char** buffer, size_t* size, const char* format)
+{
+    CSFML_CHECK_RETURN(image, sfFalse);
+
+    std::vector<sf::Uint8> output;
+    if (image->This.saveToMemory(output, format)){
+        // Create new buffer then return address and size to user
+        *buffer = reinterpret_cast<unsigned char*>(std::malloc(output.size()));
```
eXpl0it3r, line 145: Wonder if we could/should assume the buffer to already be properly prepared. 🤔 

Like from a user perspective, I wonder whether one would really pass in some raw pointers and expect the function to allocate space for it.

ChrisThrasher: I believe the usage looks like this:
```c
unsigned char* buffer;
size_t size;
sfImage_saveToFile(image, &buffer, &size, "someformatstring");
```
Given C's limitations, this was the best API I could come up with. The problem with requiring users to prepare the buffer is that I'm not sure how they're supposed to correctly size that buffer. What I wrote means users don't have to know or care what the size it beforehand.

dogunbound: I think it's fine to malloc here. Requiring the user the do a malloc would in a sense, remove the "Simple" in SFML IMHO.

dogunbound (approved):
I really tried looking for a way to save stuff to the memory in sdl2. It doesn't seem like that have a way??? Maybe I didn't look hard enough. Anyways, documentation of buffer is all that's needed for me.

kimci86:
> I really tried looking for a way to save stuff to the memory in sdl2. It doesn't seem like that have a way???

It is possible using a `SDL_RWops` object created with [SDL_RWFromMem](https://wiki.libsdl.org/SDL2/SDL_RWFromMem) and saving the image with [SDL_SaveBMP_RW](https://wiki.libsdl.org/SDL2/SDL_SaveBMP_RW). It requires to have a large enough memory buffer beforehand. I guess it is also possible to write to a dynamic buffer by using a `SDL_RWops`with custom callbacks.

dogunbound:
> > I really tried looking for a way to save stuff to the memory in sdl2. It doesn't seem like that have a way???
> 
> It is possible using a [`SDL_RWops`](https://wiki.libsdl.org/SDL2/SDL_RWops) object created with [`SDL_RWFromMem`](https://wiki.libsdl.org/SDL2/SDL_RWFromMem) and saving the image with [`SDL_SaveBMP_RW`](https://wiki.libsdl.org/SDL2/SDL_SaveBMP_RW). It requires to have a large enough memory buffer beforehand. I guess it is also possible to write to a dynamic buffer by using a `SDL_RWops` with custom callbacks.

If that's true, I think my other suggestion where you have another function malloc the buffer before using it would be most similar.

kimci86:
What about returning the vector in a C wrapper? This allows to avoid the copy of SFML's output into a new buffer. I explored the idea here: https://github.com/SFML/CSFML/compare/image_save_to_memory...kimci86:CSFML:image_save_to_memory

eXpl0it3r:
What might also be an interesting thought process is, how this can be called from other languages, e.g. how it would be implemented in SFML.Net

ChrisThrasher:
> What about returning the vector in a C wrapper? This allows to avoid the copy of SFML's output into a new buffer. I explored the idea here: [06faf7b](https://github.com/SFML/CSFML/commit/06faf7b51392991f051a4611a867b93d82b005e1)

I'm open to the idea, but I'm not in love with it since it adds quite a bit of code that we currently would only use in one place. Is there a 2nd location in CSFML or any speculated future uses of this abstraction?

src/SFML/Graphics/Image.cpp
oprypin: It's impossible for the user to predict how big the buffer should be. After all, image compression makes it so the data size can be anything whatsoever. I think this solution as suggested is the best that can be done with the given API.

oprypin: But it is true that this is the first ever usage of `malloc` in this repository (previously it was all `new`). And it's possible that the user is using yet another kind of allocator but will need to deal with `free` specifically only for this use case.

oprypin:
> What about returning the vector in a C wrapper? This allows to avoid the copy of SFML's output into a new buffer. I explored the idea here: [06faf7b](https://github.com/SFML/CSFML/commit/06faf7b51392991f051a4611a867b93d82b005e1)

This is actually a great idea, *but* it shouldn't be *returned*, it should still be passed as an output arg and instantiated by the user, then we'd also eliminate *all* copying and leave all mallocs up to the user. There's also an issue with your commit that the `new` allocator is used but the user then can't correctly `delete` the memory from C land.

As such, my suggestion would be a slight edit to your commit, then I like how it looks:
https://github.com/oprypin/CSFML/commit/e4f82f0df9080c87a347f6bcee6bdc6bcb0e3fdf

Or how it looks all together:
https://github.com/oprypin/CSFML/compare/a2c73376f96dca6ecd2034aeac4e8a4e22087830~3...e4f82f0df9080c87a347f6bcee6bdc6bcb0e3fdf
(sure would be neat if this branch was now merged as a PR authored by 3 people :innocent:)

I think with this, all the mentioned concerns about allocation are eliminated.

Maybe this new type should be moved out of Graphics module, though?

oprypin:
I am not directly using CSFML at all, but I am saying that the approach I took there is like this proposed alternative:

> Or how it looks all together:
> [oprypin/csfml@e4f82f0~3...d8281df](https://github.com/oprypin/csfml/compare/e4f82f0df9080c87a347f6bcee6bdc6bcb0e3fdf~3...d8281dfdb5c168dd291b189919ec945098145521)

ChrisThrasher:
This is one of the least CSFML 2.6.0 APIs to implement before we can release it. It's important we not stall the release for too much longer. Can we accept the PR in its current state since honestly this interface is unlikely to get used very much to begin with?

If we want to relitigate these decisions while developing CSFML 3 then so be it but I'd rather not let this PR sit for more weeks or months further stalling shipping CSFML 2.6.0.

oprypin:
We have named several disadvantages of this PR in the current state, which https://github.com/oprypin/csfml/compare/e4f82f0df9080c87a347f6bcee6bdc6bcb0e3fdf~3...d8281dfdb5c168dd291b189919ec945098145521 doesn't have. Why not merge it and discuss from there? Maybe that's even the reason that the PR is stalled - a solution was proposed but not acknowledged

ChrisThrasher:
I don't think we need something as complex as what you proposed. The reality is that we're letting a relatively unimportant PR stall an entire release. This API is not important enough to warrant this much attention. If someone wants to submit an alternative PR (with all the right docs already written and passing CI) then we can consider an alternative but so far nobody has stepped up to do that so this is our only option at the moment.

oprypin:
This method would lose most of its useful properties this way. If it's so unimportant then why was it added to SFML in the first place?

Again, let me list the deficiencies:

* The user must guess the buffer size, or probably provide a buffer that is surely too big for the destination image, or else call this method twice to get the correct size first.
* One cannot reuse an existing buffer, it must be allocated on every call, on the heap and then disposed manually.
* It is the first ever use of `malloc` in this library (adding another allocator just for this) **and the first time the user is forced to de-allocate memory specifically using C standard library.**
* Bindings to most other languages will be forced to re-copy this data in order to expose it back to the target language.

You mention "complexity" and "abstraction", but until C has anonymous tuples or something, putting data into a struct of 2 members is as simple as it gets. Plus, that struct directly mirrors the single `vector` parameter just like it is in the C++ API.

ChrisThrasher:
> The user must guess the buffer size

This is not true. The API does not require preallocating anything or having to guess any sizes beforehand.

Are you willing to write up a PR for your proposed implementation?

oprypin:
Sorry, you're right, it does not require guessing

dogunbound (approved):
As long as we at least document what is getting freed and what the user has to manage, then LGTM!!!

ChrisThrasher:
https://github.com/SFML/CSFML/blob/a80542d67d86dd37b9c6693e11ee666cbcfbe9f1/src/SFML/Window/VideoMode.cpp#L47-L69

Perhaps we copy what something like `sfVideoMode_getFullscreenModes` is doing. It internally creates a static `std::vector` and then assigns into that vector before finally returning the address of the first member of that vector and the size is an out parameter. This means so long as you don't call `sfVideoMode_getFullscreenModes` a 2nd time you can keep using the pointer that it returned.

ChrisThrasher:
```cpp
sfBool sfImage_saveToMemory(const sfImage* image, unsigned char** buffer, size_t* size, const char* format)
{
    CSFML_CHECK_RETURN(image, sfFalse);

    static std::vector<sf::Uint8> output;
    if (image->This.saveToMemory(output, format)){
        // Create new buffer then return address and size to user
        *buffer = &output[0];
        *size = output.size();
        return sfTrue;
    }

    *buffer = NULL;
    *size = 0;
    return sfFalse;
}
```
This removes the `malloc` and better matches how other CSFML APIs behave.

oprypin:
Ah no, that would just mean that you can't keep two images at once and look at them

ChrisThrasher:
I might be coming around to having an `sfBuffer` struct. Are you willing to write docs for all the new APIs you added? If so then we can merge that into this PR and discuss from there.

oprypin:
I looked into it briefly, but sorry, I'm not willing to write that boilerplate. Actually there are a lot of small but still non-trivial decisions in creating the new files for this new class

kimci86:
I would be happy to write the clean implementation with sfBuffer. I expect to have time do it by the end of the week.

ChrisThrasher:
I'd love to see that 👍🏻

kimci86:
Here we go: https://github.com/SFML/CSFML/compare/image_save_to_memory...kimci86:CSFML:image_save_to_memory

ChrisThrasher:
That's a clean solution now that I really grasp what's going on. Can't argue against that. If you open a PR I'll close this one.

ChrisThrasher:
#205 is better. Let's focus on that PR instead.

PR #200, opened 2023-07-07
Enable and fix compiler warnings

ChrisThrasher:
This caught a few bugs with incorrect conversions, string truncations, and incorrect transform comparisons.

### DIFF
--- a/src/SFML/Network/IpAddress.cpp
+++ b/src/SFML/Network/IpAddress.cpp
@@ -35,8 +35,8 @@ namespace
     // Helper function for converting a SFML address to a CSFML one
     sfIpAddress fromSFMLAddress(sf::IpAddress address)
     {
-        sfIpAddress result;
-        strncpy(result.address, address.toString().c_str(), 16);
+        sfIpAddress result = {0};
+        strncpy(result.address, address.toString().c_str(), 15);
 
         return result;
     }
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -1,0 +1,103 @@
+# from here:
+#
+# https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
+
+# Helper function to enable compiler warnings for a specific set of files
+function(set_file_warnings)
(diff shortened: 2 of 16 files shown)


src/SFML/Network/IpAddress.cpp
```diff
@@ -36,7 +36,7 @@ namespace
     sfIpAddress fromSFMLAddress(sf::IpAddress address)
     {
         sfIpAddress result;
-        strncpy(result.address, address.toString().c_str(), 16);
+        strncpy(result.address, address.toString().c_str(), 15);
```
eXpl0it3r, line 39: No `\0` terminator?

ChrisThrasher: `result.address` is a `char` array of length 16. That means it can only fit 15 character + `\0`. If you call [`strncpy`](https://en.cppreference.com/w/cpp/string/byte/strncpy) with a count of 16 and the string you're copying from with 16 characters or longer, you end up copying the first 16 non-null character into that array of 16 characters and are left with a string that is not null terminated. If the string is not null terminated, you're prone to accidentally read off the end of that array.

eXpl0it3r: Ah, I see

kimci86: Don't you need to set the last char? I believe it is not initialized.
`result.address[15] = '\0';`

ChrisThrasher: `strncpy` fills in all remaining characters will null so we don’t have to do that manually

ChrisThrasher: Good point. We may want to initialize that buffer to all zeros first.

ChrisThrasher: https://github.com/SFML/CSFML/compare/515ef0ef193f3f8332be271f0aca186316d68d3c..5bdfdb4b10d25134ad2d8b877ef519958e3f2a30

I made sure that we're initializing the `sfIpAddress` before writing to it. I wish there was a more elegant solution but this is the best I could come up with.

kimci86 (approved):

eXpl0it3r (approved):

PR #201, opened 2023-07-08
Expand CI pipeline

ChrisThrasher:
This PR started off adding new compilers. Eventually I added static builds (instead of only shared builds) and quickly discovered there were some issues with incorrect `dllimport`/`dllexport` statements. I found out that CSFML lacked the same `CSFML_STATIC` definition that SFML has for supported static library builds so I ported that over from SFML in an identical fashion and now the new and improved CI pipeline is happy.

EDIT: I removed the static library bits so this PR is now only about adding CI jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
         - { name: Windows VS2022, os: windows-2022 }
         - { name: Windows ClangCL, os: windows-2022, flags: -T ClangCL }
         - { name: Windows Clang,  os: windows-2022, flags: -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -GNinja }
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
The reason we didn't really support static linking with CSFML is, that it stops being a C-library when you link it statically, as [the C linker wouldn't know what to do with these C++ symbols](https://en.sfml-dev.org/forums/index.php?topic=13908.msg97523#msg97523). Not exactly sure if allowing to link CSFML statically as C++ library is really useful in any way, as you then could just directly use SFML to begin with... 🤔

ChrisThrasher:
If static builds aren't reasonable then let's hardcode the use of shared libraries into CSFML itself. We can use the `SHARED` argument to `add_library` to make sure only shared libs get built and totally ignore the value of `BUILD_SHARED_LIBS`.

For now I'll remove the static lib stuff from this PR so it's only about new CI jobs.

eXpl0it3r (approved):

PR #204, opened 2023-09-06
Disable faulty Windows Clang job

ChrisThrasher:
The problem is that the Windows CI image hasn't yet be updated to Clang 16. See this PR for progress on fixing it

https://github.com/actions/runner-images/pull/8134

This has been merged but we still have to wait for the changes to propogate to all runners.


ChrisThrasher:
https://github.com/SFML/CSFML/actions/runs/6103425060

Here's an example of this job failing.

PR #205, opened 2023-09-07
Add `sfBuffer` wrapper and implement `sf::Image::saveToMemory`

kimci86:
Alternative to #195 using a C wrapper over an `std::vector`.

### DIFF
--- a/src/SFML/Graphics/Image.cpp
+++ b/src/SFML/Graphics/Image.cpp
@@ -27,8 +27,10 @@
 ////////////////////////////////////////////////////////////
 #include <SFML/Graphics/Image.h>
 #include <SFML/Graphics/ImageStruct.h>
+#include <SFML/System/BufferStruct.h>
 #include <SFML/Internal.h>
 #include <SFML/CallbackStream.h>
+
 
 ////////////////////////////////////////////////////////////
 sfImage* sfImage_create(unsigned int width, unsigned int height)
@@ -128,6 +130,14 @@ void sfImage_destroy(sfImage* image)
 sfBool sfImage_saveToFile(const sfImage* image, const char* filename)
 {
     CSFML_CALL_RETURN(image, saveToFile(filename), sfFalse);
+}
+
+
+////////////////////////////////////////////////////////////
+sfBool sfImage_saveToMemory(const sfImage* image, sfBuffer* output, const char* format)
+{
+    CSFML_CHECK_RETURN(output, sfFalse);
+    CSFML_CALL_RETURN(image, saveToMemory(output->buffer, format), sfFalse);
 }
 
 
--- a/include/SFML/Graphics/Image.h
+++ b/include/SFML/Graphics/Image.h
@@ -32,6 +32,7 @@
 #include <SFML/Graphics/Color.h>
 #include <SFML/Graphics/Rect.h>
 #include <SFML/Graphics/Types.h>
+#include <SFML/System/Buffer.h>
 #include <SFML/System/InputStream.h>
 #include <SFML/System/Vector2.h>
 #include <stddef.h>
@@ -156,8 +157,29 @@ CSFML_GRAPHICS_API void sfImage_destroy(sfImage* image);
 ///
 /// \return sfTrue if saving was successful
 ///
+/// \see sfImage_saveToMemory
+///
 ////////////////////////////////////////////////////////////
 CSFML_GRAPHICS_API sfBool sfImage_saveToFile(const sfImage* image, const char* filename);
+
+////////////////////////////////////////////////////////////
+/// \brief Save the image to a buffer in memory
+///
+/// The format of the image must be specified.
+/// The supported image formats are bmp, png, tga and jpg.
+/// This function fails if the image is empty, or if
+/// the format was invalid.
+///
+/// \param image  Image object
+/// \param output Buffer to fill with encoded data
+/// \param format Encoding format to use
+///
+/// \return sfTrue if saving was successful
+///
+/// \see sfImage_saveToFile
+///
+////////////////////////////////////////////////////////////
+CSFML_GRAPHICS_API sfBool sfImage_saveToMemory(const sfImage* image, sfBuffer* output, const char* format);
 
 ////////////////////////////////////////////////////////////
 /// \brief Return the size of an image
--- a/include/SFML/System.h
+++ b/include/SFML/System.h
@@ -30,6 +30,7 @@
 ////////////////////////////////////////////////////////////
 
 #include <SFML/Config.h>
+#include <SFML/System/Buffer.h>
 #include <SFML/System/Clock.h>
 #include <SFML/System/InputStream.h>
 #include <SFML/System/Mutex.h>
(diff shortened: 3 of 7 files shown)


src/SFML/Graphics/Buffer.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CSFML_CHECK_RETURN(buffer, 0);
+    return buffer->vector.size();
+}
+
+
+////////////////////////////////////////////////////////////
+CSFML_GRAPHICS_API const sfUint8* sfBuffer_getData(const sfBuffer* buffer)
+{
+    CSFML_CHECK_RETURN(buffer, NULL);
+    return buffer->vector.data();
```
ChrisThrasher, line 59: Do we want to `assert(buffer->vector.size() > 1)`? The standard specifies that if the vector is empty then `.data()` may or may not return a null pointer. I suppose it's not necessarily a bug to call it under those circumstances but it's worth consideration.

src/SFML/Graphics/Image.cpp
```diff
@@ -131,6 +133,14 @@ sfBool sfImage_saveToFile(const sfImage* image, const char* filename)
 }
 
 
+////////////////////////////////////////////////////////////
+sfBool sfImage_saveToMemory(const sfImage* image, sfBuffer* output, const char* format)
+{
+    CSFML_CHECK_RETURN(output, sfFalse);
+    CSFML_CALL_RETURN(image, saveToMemory(output->vector, format), sfFalse);
```
ChrisThrasher, line 140: `format` gets converted to `std::string` and it's UB to construct a string from a null pointer. Should we add a `CSFML_CHECK_RETURN` call for `format` being null?

ChrisThrasher (approved):
I appreciate you putting this together with all the right docs and everything.

src/SFML/Graphics/Buffer.cpp
kimci86: The valid range for any vector being [data(), data() + size()), the valid range for an empty vector is empty. I don't think we need a special case. It is up to the user not to dereference values out of the valid range.

src/SFML/Graphics/Image.cpp
kimci86: Nice catch, will do.

ChrisThrasher: That's fine by me

include/SFML/Graphics/Buffer.h
```diff
@@ -0,0 +1,73 @@
...
+#include <stddef.h>
+
+
+////////////////////////////////////////////////////////////
+/// \brief Create an empty buffer
+///
+/// \return A new sfBuffer object, or NULL if it failed
+///
+////////////////////////////////////////////////////////////
+CSFML_GRAPHICS_API sfBuffer* sfBuffer_create();
```
kimci86, line 42: I forgot the `void` parameter, will fix.

src/SFML/Graphics/Buffer.cpp
```diff
@@ -0,0 +1,60 @@
...
```
eXpl0it3r, line 59: `data()` is C++11, I guess we should keep CSFML 2.6 aligned with the C++ versioning if possible

src/SFML/Graphics/BufferStruct.h
```diff
@@ -0,0 +1,44 @@
...
+#include <SFML/Config.hpp>
+#include <vector>
+
+
+////////////////////////////////////////////////////////////
+// Internal structure of sfBuffer
+////////////////////////////////////////////////////////////
+struct sfBuffer
+{
+    std::vector<sf::Uint8> vector;
```
eXpl0it3r, line 40: I would call the member after the container type, maybe just `buffer` instead?

include/SFML/Graphics/Buffer.h
eXpl0it3r, line 1: Kind of wondering whether this should live in the System part, but I guess as long as we just use it in the Graphics module, lets keep it here.

eXpl0it3r (changes requested):

src/SFML/Graphics/Buffer.cpp
kimci86: Well spotted! Fixed.

src/SFML/Graphics/BufferStruct.h
kimci86: Ok, done

kimci86:
I used the following code to test the API.
It made me notice the `#include <SFML/Graphics/Buffer.h>` was missing from `SFML/Graphics.h` file and fixed it.

```c
#include <SFML/Graphics.h>
#include <stdio.h>

int main(void)
{
    const int WIDTH = 256;
    sfImage* image = sfImage_createFromColor(WIDTH, 1, sfMagenta);
    for (int x = 0; x < WIDTH; ++x)
        sfImage_setPixel(image, x, 0, sfColor_fromRGB(x, 255, x));

    sfBuffer* buffer = sfBuffer_create();
    sfBool success = sfImage_saveToMemory(image, buffer, "png");
    printf("Succes: %d\n", success);

    sfImage_destroy(image);

    FILE* f = fopen("out.png", "wb");
    fwrite(sfBuffer_getData(buffer), 1, sfBuffer_getSize(buffer), f);
    fclose(f);

    sfBuffer_destroy(buffer);

    return 0;
}
```

eXpl0it3r (approved):

eXpl0it3r:
@oprypin any comment from your side? 🙂

oprypin:
It's good.
One thing is: putting `sfBuffer` into `graphics` module is strange because then we can't add a `sfBuffer` into `system` or `window` if we need it later. But it may be also strange to put it into `system` right away.

eXpl0it3r:
Hmmm I put it off as "we can always move it later", but then we'd be breaking code. While I don't really hold myself to the same non-breaking standard for CSFML as for SFML, I think it might after all be beneficial to reduce it where possible.

So I think, it does after all matter and we should move `sfBuffer` to the system module. So it could be reused e.g. in the audio module in the future, without "breaking" the API.

ChrisThrasher:
There are no other `saveToMemory` functions in SFML and if we added more they'd be added in SFML 3 and thus CSFML 3 so we can get away with leaving it where it is if we want. I'm fine moving it to System though. Either is fine with me.

oprypin:
Hm but wait, if `sfBuffer` is first in `graphics` but then is moved to `system`, can't it just be re-exposed in `graphics` and then not break anything?

oprypin:
But anyway probably `system` is better.  Sorry to cause you work :sweat_smile:

eXpl0it3r:
> Hm but wait, if `sfBuffer` is first in `graphics` but then is moved to `system`, can't it just be re-exposed in `graphics` and then not break anything?

It could, but I wouldn't want such a workaround, just to avoid some potential breaking code. No need to add technical debt.

kimci86:
I moved sfBuffer into csfml-system.

oprypin (approved):

eXpl0it3r (approved):
Thank you ❤️

PR #215, opened 2023-12-20
Add example program

### LINKED ISSUE #211: Add example programs
Having example projects that link to CSFML can help catch linker issues that are not detectable when only building the library itself.

ChrisThrasher:
Closes #211 

I had to fix some function prototypes that were using `()` instead of `(void)` to list their parameters. The latter is required when compiling these headers as C code instead of C++ code. This example program is the first time internally to CSFML that CSFML headers were compiled as C code instead of C++, thus uncovering this problem.


examples/example.c
```diff
@@ -0,0 +1,74 @@
...
+    /* Create the main window */
+    const sfVideoMode mode = {800, 600, 32};
+    sfRenderWindow* window = sfRenderWindow_create(mode, "SFML window", sfResize | sfClose, NULL);
+    if (!window)
+        return EXIT_FAILURE;
+
+    /* Load a sprite to display */
+    sfTexture* texture = sfTexture_createFromFile("sfml_logo.png", NULL);
+    if (!texture)
+        return EXIT_FAILURE;
```
kimci86, line 17:
```suggestion
    if (!texture)
    {
        sfRenderWindow_destroy(window);
        return EXIT_FAILURE;
    }
```

examples/example.c
```diff
@@ -0,0 +1,74 @@
...
+    sfSprite* sprite = sfSprite_create();
+    sfSprite_setTexture(sprite, texture, sfTrue);
+    const sfVector2f spritePosition = {200, 200};
+    sfSprite_setPosition(sprite, spritePosition);
+
+    /* Create a graphical text to display */
+    sfFont* font = sfFont_createFromFile("tuffy.ttf");
+    if (!font)
+        return EXIT_FAILURE;
```
kimci86, line 26:
```suggestion
    if (!font)
    {
        sfSprite_destroy(sprite);
        sfTexture_destroy(texture);
        sfRenderWindow_destroy(window);
        return EXIT_FAILURE;
    }
```

examples/example.c
```diff
@@ -0,0 +1,74 @@
...
+    sfText* text = sfText_create();
+    sfText_setString(text, "Hello, SFML!");
+    sfText_setFont(text, font);
+    sfText_setCharacterSize(text, 50);
+
+    /* Load a music to play */
+    sfMusic* music = sfMusic_createFromFile("doodle_pop.ogg");
+    if (!music)
+        return EXIT_FAILURE;
```
kimci86, line 35:
```suggestion
    if (!music)
    {
        sfText_destroy(text);
        sfFont_destroy(font);
        sfSprite_destroy(sprite);
        sfTexture_destroy(texture);
        sfRenderWindow_destroy(window);
        return EXIT_FAILURE;
    }
```

kimci86 (commented):
Maybe also cleanup resources when loading fails?

examples/example.c
```diff
@@ -0,0 +1,74 @@
+#include <SFML/Audio.h>
+#include <SFML/Graphics.h>
+
+#include <stdlib.h>
+
+int main(void)
+{
+    /* Create the main window */
```
eXpl0it3r, line 8: Maybe we can use single line comments

eXpl0it3r:
Do you want to enable it in CI or are you gonna do that with the other PR?

ChrisThrasher:
> Do you want to enable it in CI or are you gonna do that with the other PR?

My bad that's fixed now

examples/example.c
```diff
@@ -0,0 +1,95 @@
...
+#include <stdlib.h>
+
+int main(void)
+{
+    // Create the main window
+    const sfVideoMode mode = {800, 600, 32};
+    sfRenderWindow* window = sfRenderWindow_create(mode, "SFML window", sfResize | sfClose, NULL);
+    if (!window)
+    {
+        sfRenderWindow_destroy(window);
```
kimci86, line 13: There is no window to destroy at this point. We do not need to call `sfRenderWindow_destroy(NULL)`.
Similar remark applies to other error management if statements below.

kimci86 (approved):

PR #218, opened 2023-12-20
Assorted cleanup

ChrisThrasher:
See each commit for individual changes. These are all non-functional changes that make boring internal details a little more uniform or well-written.


eXpl0it3r (approved):

tools/nuget/README.md
```diff
@@ -11,11 +11,11 @@ All build scripts follow the same routine:
 
 
 * `build.win.ps1`
-    * A [PowerShell](https://github.com/PowerShell/PowerShell) script to build Windows native libraries. By default, both `win-x86` and `win-x64` are built. To build only one specific Runtime Identifier, pass it to the script as a parameter (`./build.win.ps1 win-x64`)
```
kimci86, line 14: Two trailing spaces is a line break in markdown: https://daringfireball.net/projects/markdown/syntax#p

ChrisThrasher: There's already an empty newline after this. Wouldn't that achieve the same thing?

eXpl0it3r: It's the difference between:

```md
Hello  
World
```

Hello  
World

```md
Hello
World
```

Hello
World

As you can see, one respects the line break, while the other doesn't.

eXpl0it3r: It's a line wrap within the paragraph. Usually if you have just a line break, the Markdown will be rendered as one contiguous line. When you add two spaces, the lines will be wrapped.

Think of the difference between a line wrap within a paragraph (ctrl+enter) and a new paragraph section (enter). For HTML that's the difference between `<br>` and a new `<p>`.

I think GitHub doesn't always follow that rule.

ChrisThrasher: I rolled back the changes to this file

eXpl0it3r (approved):

PR #220, opened 2023-12-29
Convert private C++ headers to .hpp file extension

### LINKED ISSUE #208: Use .hpp extension for private C++ headers
All of the headers in src/ are actually C++ code since they use C++ headers and C++ symbols. We should use a .hpp extension for such files instead of .h.

ChrisThrasher:
Closes #208 

These are C++ source code files. To be consistent with SFML they should use an .hpp extension. This also has the benefit of making it easier to readers to understand whether a given file contains C or C++ code.


eXpl0it3r:
This makes it really tricky to differentiate between CSFML headers and SFML headers next to each other.

Like this:
image

ChrisThrasher:
CSFML ought to use a `CSFML` prefix for its headers. I think that's a better solution that using an ambiguous file extension. I'm aware we can't do this in CSFML 2 but we can do it in v3 and we can do it for the private headers.

eXpl0it3r:
I guess for CSFML 3 it's okay to move everything to `CSFML/`

ChrisThrasher:
> I guess for CSFML 3 it's okay to move everything to `CSFML/`

https://github.com/SFML/CSFML/compare/master...include_dir

Here's a PR that changes the SFML/ directory prefix to CSFML/. If you like this I'll submit a PR.

ChrisThrasher:
Updated this PR to be up to date with `master`. Currently this PR depends on #266 and #269. Once those are merged then this will be ready for review.

eXpl0it3r (approved):

PR #225, opened 2024-02-04
Begin work on CSFML v3

ChrisThrasher:
With 2.6.0 now released we're ready to start working towards v3! With SFML 3 not even finalized it will take a while to get there but luckily there are some easy places to start. This branch includes a mixture of breakages that could not happen in the v2 release series and modernizations that require newer language standards than what v2 uses.

To match SFML this PR upgrades us to CMake 3.22 and C++17. I also upgraded the C requirement to C11. I'm most happy with how requiring C99 now gives us access to basic quality of life features like fixed-width integers and boolean literals. CSFML is being brought into the 21st century 😊

I'm no expert in C standards but from what I've heard C17 is basically identical to C11. It probably wouldn't hurt to eventually raise the C requirement to C17 but for now I picked C11 to ensure we have access to newer features that may prove useful.


src/SFML/Graphics/Shader.cpp
```diff
@@ -456,19 +393,19 @@ unsigned int sfShader_getNativeHandle(const sfShader* shader)
...
+    return sf::Shader::isAvailable();
 }
 
 
 ////////////////////////////////////////////////////////////
-sfBool sfShader_isGeometryAvailable(void)
+bool sfShader_isGeometryAvailable(void)
 {
-    return sf::Shader::isGeometryAvailable() ? sfTrue : sfFalse;
+    return sf::Shader::isGeometryAvailable();
```
eXpl0it3r, line 410: Does C++ `bool` implicitly convert to C `bool`?

src/SFML/Graphics/Shape.cpp
```diff
@@ -167,9 +167,9 @@ sfTransform sfShape_getInverseTransform(const sfShape* shape)
 
 
 ////////////////////////////////////////////////////////////
-void sfShape_setTexture(sfShape* shape, const sfTexture* texture, sfBool resetRect)
+void sfShape_setTexture(sfShape* shape, const sfTexture* texture, bool resetRect)
 {
-    CSFML_CALL(shape, setTexture(texture ? texture->This : NULL, resetRect == sfTrue));
+    CSFML_CALL(shape, setTexture(texture ? texture->This : nullptr, resetRect == true));
```
eXpl0it3r, line 172: Do we still need the `resetRect == true` or can we just pass down `resetRect`?

src/SFML/Window/VideoMode.cpp
```diff
@@ -52,26 +52,26 @@ const sfVideoMode* sfVideoMode_getFullscreenModes(size_t* count)
     if (modes.empty())
     {
         const std::vector<sf::VideoMode>& SFMLModes = sf::VideoMode::getFullscreenModes();
-        for (std::vector<sf::VideoMode>::const_iterator it = SFMLModes.begin(); it != SFMLModes.end(); ++it)
+        for (const auto& SFMLMode : SFMLModes)
```
eXpl0it3r, line 55: I'd rename `SFMLModes` and `SFMLMode` to `sfmlModes` and `sfmlMode`

eXpl0it3r (changes requested):
Just some bike shedding.

If the `var == true` aren't needed anymore, I'd want to see them removed.

Other than that, I think it can be merged. I've already created a `2.6.x` branch 🙂

src/SFML/Graphics/Shader.cpp
ChrisThrasher: Yes it does. As far as I'm aware they're indistinguishable.

ChrisThrasher:
Good feedback. I think I've addressed all your comments.

ChrisThrasher:
I found some more modernizations/simplifications to make and am done with that effort now. This PR is ready for review again. Better merge it soon before I find more things to change :)

kimci86 (approved):
Nothing else to complain about when skimming through :+1:

Marioalexsan (approved):

PR #229, opened 2024-02-11
Add count parameter to sfVulkan_getGraphicsRequiredInstanceExtensions

### LINKED ISSUE #227: Insufficient documentation for `sfVulkan_getGraphicsRequiredInstanceExtensions`?
I'm working on updating the [Ada binding](https://github.com/mgrojo/ASFML) to version 2.6.0. There's a particular function that I don't know how to interface to. Forgive me if it's due to a lack of C/C++ knowledge on my part, but in any case, better documentation on how to use this function would be appreciated:

```c
/// \brief Get Vulkan instance extensions required for graphics
///
/// \return Vulkan instance extensions required for graphics
///
CSFML_WINDOW_API const char* const* sfVulkan_getGraphicsRequiredInstanceExtensions(void);
```
I suppose that it returns a list of strings, but how is the caller aware of how many of them are available? For v2.5 I had to interface to a similar case, and the number of strings was returned through a `count` parameter. Compare with this case:

```c

/// \brief Get a list of the names of all availabe audio capture devices
///
/// This function returns an array of strings (null terminated),
/// containing the names of all availabe audio capture devices.
/// If no devices are available then NULL is returned.
///
/// \param count Pointer to a variable that will be filled with the number of modes in the array
///
/// \return An array of strings containing the names
///
CSFML_AUDIO_API const char** sfSoundRecorder_getAvailableDevices(size_t* count);
```

mgrojo:
Based on `sfSoundRecorder_getAvailableDevices()`

Fixes #227

### DIFF
--- a/include/SFML/Window/Vulkan.h
+++ b/include/SFML/Window/Vulkan.h
@@ -26,6 +26,7 @@
 // Headers
 ////////////////////////////////////////////////////////////
 #include <SFML/Window/Export.h>
+#include <stddef.h>
 
 
 typedef struct VkInstance_T* VkInstance;
@@ -76,7 +77,9 @@ CSFML_WINDOW_API sfVulkanFunctionPointer sfVulkan_getFunction(const char* name);
 ////////////////////////////////////////////////////////////
 /// \brief Get Vulkan instance extensions required for graphics
 ///
+/// \param count Pointer to a variable that will be filled with the number of modes in the array
+///
 /// \return Vulkan instance extensions required for graphics
 ///
 ////////////////////////////////////////////////////////////
-CSFML_WINDOW_API const char* const* sfVulkan_getGraphicsRequiredInstanceExtensions(void);
+CSFML_WINDOW_API const char* const* sfVulkan_getGraphicsRequiredInstanceExtensions(size_t* count);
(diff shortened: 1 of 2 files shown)


include/SFML/Window/Vulkan.h
```diff
@@ -76,7 +77,9 @@ CSFML_WINDOW_API sfVulkanFunctionPointer sfVulkan_getFunction(const char* name);
 ////////////////////////////////////////////////////////////
 /// \brief Get Vulkan instance extensions required for graphics
 ///
+/// \param count Pointer to a variable that will be filled with the number of modes in the array
```
kimci86, line 80:
```suggestion
/// \param count Pointer to a variable that will be filled with the number of instance extensions in the array
```

mgrojo: I blindly copied that from `sfSoundRecorder_getAvailableDevices`. It's strange, because "modes" is also meaningless there (I think), so maybe it was also blindly copied from elsewhere :smile:

Marioalexsan (approved):

ChrisThrasher (approved):
Thanks for fixing this!

PR #236, opened 2024-02-27
Implement missing functions for 2.6.x

### LINKED ISSUE #235: Some CSFML 2.6.x API Discrepancies
I've noticed these while trying to update SFML.Net to 2.6.0.

- Mouse is missing position getters / setters for WindowBase (i.e. `sfMouse_getPositionWindowBase` and `sfMouse_setPositionWindowBase`)
- Touch is missing position getter for WindowBase (i.e. `sfTouch_getPositionWindowBase`)
- While it might not be very useful, createVulkanSurface overloads for derived windows are missing (i.e. `sfRenderWindow_createVulkanSurface` and `sfWindow_createVulkanSurface`)
- `sfKeyboard_getDescription` returns an owning const char* that must be freed using C's free, however CSFML doesn't expose something like `sfFree` / `sfMalloc` that can be easily called from the C# side, which causes issues with memory leaks

Marioalexsan:
Figured I might as well make an attempt at this. Not sure if `sfMalloc` is needed but I added it in for completeness.

Fixes #235.

### DIFF
--- a/include/SFML/System/Alloc.h
+++ b/include/SFML/System/Alloc.h
@@ -22,46 +22,28 @@
 //
 ////////////////////////////////////////////////////////////
 
+#ifndef SFML_ALLOC_H
+#define SFML_ALLOC_H
+
 ////////////////////////////////////////////////////////////
 // Headers
 ////////////////////////////////////////////////////////////
-#include <SFML/Window/Touch.h>
-#include <SFML/Window/Touch.hpp>
-#include <SFML/Window/WindowBaseStruct.h>
-#include <SFML/Window/WindowStruct.h>
-#include <SFML/Internal.h>
+#include <SFML/System/Export.h>
 
 
 ////////////////////////////////////////////////////////////
-sfBool sfTouch_isDown(unsigned int finger)
-{
-    return sf::Touch::isDown(finger);
-}
-
+/// \brief Deallocates memory
+///
+/// This function deallocates the memory being pointed to
+/// using the free function from the C standard library.
+/// 
+/// The memory must have been previously allocated using a call
+/// to malloc.
+/// 
+/// \param ptr Pointer to the memory to deallocate 
+///
 ////////////////////////////////////////////////////////////
-sfVector2i sfTouch_getPosition(unsigned int finger, const sfWindow* relativeTo)
-{
-    sf::Vector2i sfmlPosition;
+CSFML_SYSTEM_API void sfFree(void* ptr);
 
-    if (relativeTo)
-        sfmlPosition = sf::Touch::getPosition(finger, relativeTo->This);
-    else
-        sfmlPosition = sf::Touch::getPosition(finger);
 
-    sfVector2i position = {sfmlPosition.x, sfmlPosition.y};
-    return position;
-}
-
-////////////////////////////////////////////////////////////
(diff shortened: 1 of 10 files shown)


src/SFML/System/Alloc.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    return malloc(size);
+}
+
+
+////////////////////////////////////////////////////////////
+void sfFree(void* ptr)
+{
+    free(ptr);
+}
```
ChrisThrasher, line 44: Why do we need to provide functions that wrap existing C standard library functions? Why can't users just call `free` themselves?

Marioalexsan: C# for example doesn't have access to the C standard library functions (as far as I'm aware). There is `Marshal.AllocHGlobal` and `Marshal.FreeHGlobal` but I don't think they are the same as `malloc` and `free`. Therefore, unless CSFML exports a method to free the pointers that can be P/Invoked from C#, there's no way for the C# bindings to use methods like `sfKeyboard_getDescription` without memory leaks.

ChrisThrasher: That's fair. If our API provides an allocation, we must provide a way to deallocate and if users can't use the C standard library then I suppose we're forced to add `sfFree`. Let's remove `sfMalloc` though since users should have no reasons to use that and we don't need it internally. We add it later should the need arise.

Marioalexsan: Sound good, I've pushed an update with `sfMalloc` removed.

include/SFML/System/Alloc.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef SFML_ALLOC_H
+#define SFML_ALLOC_H
+
+////////////////////////////////////////////////////////////
+// Headers
+////////////////////////////////////////////////////////////
+#include <SFML/System/Export.h>
+
+#include <stddef.h>
+
```
ChrisThrasher, line 34:
```suggestion
```

ChrisThrasher (changes requested):

include/SFML/System/Alloc.h
Marioalexsan: Should be fixed now.

src/SFML/Window/Mouse.cpp
```diff
@@ -60,3 +61,26 @@ void sfMouse_setPosition(sfVector2i position, const sfWindow* relativeTo)
...
+    sf::Vector2i sfmlPos;
+    if (relativeTo)
+        sfmlPos = sf::Mouse::getPosition(relativeTo->This);
+    else
+        sfmlPos = sf::Mouse::getPosition();
+
+    sfVector2i position = { sfmlPos.x, sfmlPos.y };
+    return position;
+}
+
```
ChrisThrasher, line 78:
```suggestion


```

ChrisThrasher (approved):
Barring an ultra-critical whitespace error I think this is good. Eager to see you fold this into SFML.Net to confirm it's all working!

ChrisThrasher:
It appears that in trying to fix this whitespace error myself then squashing the new commit that was created, I accidentally pushed these changes straight to 2.6.x. I didn't realize your PR was opened from your own repo's 2.6.x branch. I'm used to that not being the case and thus I can edit the PR's branch without affecting the target branch. Oh well. I was already going to merge it anyways so the outcome is all the same.

PR #80, opened 2015-05-19
fix rpath for OS X

vonagam:
almost copy-paste ( without framework stuff ) from SFML.

PR #243, opened 2024-03-29
Add unit testing infrastructure and initial unit tests

ChrisThrasher:
Waiting on #242 

For the most part this PR copies over the same testing infrastructure that SFML already uses. The unit tests themselves are written in C++ because C++ provides more options for writing expressive tests plus it means more in similar with SFML's tests which makes for easier maintenance. Writing our tests in pure C would not be particularly useful.

I had to make a few changes to how rpaths are treated to get the tests to work. CMake's default handling of rpaths seems to work just fine but the custom settings we were using were breaking things. I'm not sure why `CMAKE_SKIP_BUILD_RPATH` was enabled but I feel comfortable removing that. Build-time rpaths are useful. I'm also not sure why `BUILD_WITH_INSTALL_RPATH` was enabled for macOS dylibs but I also had to remove that to get the tests to work on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,6 @@ else()
 endif()
 csfml_set_option(CSFML_LINK_SFML_STATICALLY ${LINK_STATICALLY_DEFAULT} BOOL "TRUE to link to a static version of SFML, FALSE to link dynamically")
 
-# disable the rpath stuff
-set(CMAKE_SKIP_BUILD_RPATH TRUE)
-
 # define an option for choosing between static and dynamic C runtime (Windows only)
 if(SFML_OS_WINDOWS)
     set(STATIC_STD_LIBS FALSE CACHE BOOL "TRUE to statically link to the standard libraries, FALSE to use them as DLLs")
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -33,9 +33,6 @@ else()
 endif()
 csfml_set_option(CSFML_LINK_SFML_STATICALLY ${LINK_STATICALLY_DEFAULT} BOOL "TRUE to link to a static version of SFML, FALSE to link dynamically")
 
-# disable the rpath stuff
-set(CMAKE_SKIP_BUILD_RPATH TRUE)
```
ChrisThrasher, line 37: https://github.com/SFML/SFML/commit/a991fe8e4db5104133ff3cefd7ec8c38339dfeae#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR44

I tried to track down when this was added and got as far back as to SFML's original CMake support added 14 years ago. I can't find any rationale for this line which makes more comfortable removing it.

cmake/Macros.cmake
```diff
@@ -66,9 +66,7 @@ macro(csfml_add_library target)
     # build dylibs
     if(SFML_OS_MACOSX AND BUILD_SHARED_LIBS)
         # adapt install directory to allow distributing dylibs in user’s application bundle
-        set_target_properties(${target} PROPERTIES
-                              BUILD_WITH_INSTALL_RPATH 1
```
ChrisThrasher, line 70: https://github.com/SFML/CSFML/pull/80

Here's the commit that added this. No rationale was provided in the PR thread so I'm not sure why this was needed. CMake's rpath support has evolved over the years so it's possible this was required 9 years ago but is not required anymore now that we're using CMake >3.22.

eXpl0it3r (approved):

cmake/Macros.cmake
eXpl0it3r: We can always adjust when we find out what it was used for 😄

PR #245, opened 2024-04-15
Add `sfJoystick` tests

ChrisThrasher:
Most Window module APIs require a display which means pulling in the same `xvfb` infrastructure SFML uses. We can do that eventually, but for now I wrote some quick `sfJoystick` tests since these don't require a display to be ran in CI.

This is yet another instance where having CSFML live inside SFML/SFML would be useful since it could easily pick up the existing infrastructure SFML has like this `xvfb` stuff or our clang-format and clang-tidy infrastructure. This is certainly out of scope for SFML 3 but it's still something I'm considering for the future since I think it's both feasible and beneficial.


ChrisThrasher:
DLL issues are causing CI to fail. We need to make sure all of the SFML and CSFML DLLs exist in the same directory as the test executables but for some reason that is not the case at the moment.

PR #246, opened 2024-05-07
Fix Doxygen Generation

### LINKED ISSUE #210: Generated CSFML documentation does not document any functions
The generated CSFML documentation doesn't actually document any functions. It only seems to document structures which you can see under the "Classes" tab.

This may be related to https://github.com/SFML/SFML/issues/2813. I would expect the functions to appear under the file list, but if the file list is broken then maybe it's not surprising that no functions from those files are documented. In SFML it's not too bad though because the class documentation is still there.

### Some pictures

The black links are not clickable.
image

image

Structures are documented but not the functions associated with them:
image

eXpl0it3r:
Similar to https://github.com/SFML/SFML/pull/2986 this fixes a bunch of things:

- Handles generation issues with Doxygen 1.10.0
- Brings the Doxygen file up-to-date
- Fix missing function documentation, fixes #210 
- Enables the search similar to https://github.com/SFML/SFML/pull/2173 & https://github.com/SFML/SFML/pull/2477
- Fixes some URLs
- Fixes XHTML issues reported in https://github.com/SFML/SFML/issues/2812
- Fixes errors for the cookie.js (copied from https://github.com/SFML/SFML/pull/2986)
- Makes the new clipboard feature work properly (copied from https://github.com/SFML/SFML/pull/2986)

For the missing function documentation, I'm not entirely sure if the `EXTRACT_ALL` is really enough.  
Yes, you can now find the functions, e.g. by going to a file or search for it, but there's no nice overview, and unfortunately, doxygen doesn't seem to have a "function" tab and in a C source the "class" tab is only partially useful.  
I wonder, if it would still make sense to create modules to collect all the functions of the same SFML class in one overview page.

For now however, I believe this should be good enough.


ChrisThrasher (approved):

kimci86:
> but there's no nice overview, and unfortunately, doxygen doesn't seem to have a "function" tab

It is a maybe little hidden but there is a list in Files > File Members > Functions

Capture d’écran du 2024-05-08 20-11-20

kimci86 (commented):
Little suggestions: https://github.com/SFML/CSFML/compare/bugfix/doxygen-docs...kimci86:CSFML:bugfix/doxygen-docs

kimci86 (approved):

PR #247, opened 2024-05-08
Catch up with SFML 3 API changes

ChrisThrasher:
Since we last updated CSFML, SFML has changed the `sf::Event` and Audio APIs.

### DIFF
--- a/src/SFML/Audio/SoundBuffer.cpp
+++ b/src/SFML/Audio/SoundBuffer.cpp
@@ -80,11 +80,19 @@ sfSoundBuffer* sfSoundBuffer_createFromStream(sfInputStream* stream)
 
 
 ////////////////////////////////////////////////////////////
-sfSoundBuffer* sfSoundBuffer_createFromSamples(const int16_t* samples, uint64_t sampleCount, unsigned int channelCount, unsigned int sampleRate)
+sfSoundBuffer* sfSoundBuffer_createFromSamples(const int16_t*    samples,
+                                               uint64_t          sampleCount,
+                                               unsigned int      channelCount,
+                                               unsigned int      sampleRate,
+                                               sfSoundChannel*   channelMapData,
+                                               size_t            channelMapSize)
 {
     sfSoundBuffer* buffer = new sfSoundBuffer;
 
-    if (!buffer->This.loadFromSamples(samples, sampleCount, channelCount, sampleRate))
+    std::vector<sf::SoundChannel> channelMap(channelMapSize);
+    for (std::size_t i = 0; i < channelMap.size(); ++i)
+        channelMap[i] = static_cast<sf::SoundChannel>(channelMapData[i]);
+    if (!buffer->This.loadFromSamples(samples, sampleCount, channelCount, sampleRate, channelMap))
     {
         delete buffer;
         buffer = nullptr;
--- a/src/SFML/Audio/SoundRecorder.cpp
+++ b/src/SFML/Audio/SoundRecorder.cpp
@@ -76,13 +76,6 @@ bool sfSoundRecorder_isAvailable(void)
 
 
 ////////////////////////////////////////////////////////////
-void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval)
-{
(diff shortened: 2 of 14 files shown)


src/SFML/ConvertEvent.h
```diff
@@ -37,84 +37,122 @@
...
+    {
+        event->type         = sfEvtKeyReleased;
+        event->key.code     = static_cast<sfKeyCode>(keyReleased->code);
+        event->key.scancode = static_cast<sfScancode>(keyReleased->scancode);
+        event->key.alt      = keyReleased->alt;
+        event->key.control  = keyReleased->control;
+        event->key.shift    = keyReleased->shift;
+        event->key.system   = keyReleased->system;
+    }
+    else if (const auto* keyPressed = sfmlEvent.getIf<sf::Event::KeyReleased>())
```
kimci86, line 61:
```suggestion
    else if (const auto* keyPressed = sfmlEvent.getIf<sf::Event::KeyPressed>())
```

src/SFML/ConvertEvent.h
kimci86, line 1: Missing unit types: `Closed`, `FocusLost`, `FocusGained`, `MouseEntered`, `MouseLeft`.

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -80,11 +80,19 @@ sfSoundBuffer* sfSoundBuffer_createFromStream(sfInputStream* stream)
 
 
 ////////////////////////////////////////////////////////////
-sfSoundBuffer* sfSoundBuffer_createFromSamples(const int16_t* samples, uint64_t sampleCount, unsigned int channelCount, unsigned int sampleRate)
+sfSoundBuffer* sfSoundBuffer_createFromSamples(const int16_t*    samples,
```
kimci86, line 83: Corresponding declaration was not updated.

src/SFML/Audio/SoundRecorder.cpp
```diff
@@ -75,13 +75,6 @@ bool sfSoundRecorder_isAvailable(void)
 }
 
 
-////////////////////////////////////////////////////////////
-void sfSoundRecorder_setProcessingInterval(sfSoundRecorder* soundRecorder, sfTime interval)
```
kimci86, line 79: Corresponding declaration was not removed.

src/SFML/Audio/SoundStream.cpp
```diff
@@ -35,9 +35,11 @@ sfSoundStream* sfSoundStream_create(sfSoundStreamGetDataCallback onGetData,
                                     sfSoundStreamSeekCallback    onSeek,
                                     unsigned int                 channelCount,
                                     unsigned int                 sampleRate,
-                                    void*                        userData)
+                                    void*                        userData,
+                                    sfSoundChannel*              channelMapData,
+                                    size_t                       channelMapSize)
```
kimci86, line 40: Corresponding declaration was not updated.
Also I am wondering if maybe `userData` should remain the last parameter.

src/SFML/Audio/SoundBuffer.cpp
ChrisThrasher: I'm used to C++ where changing a member function without updating the declaration is a compiler error but alas these are not member functions.

src/SFML/ConvertEvent.h
ChrisThrasher: These were missing before. I wonder if that was intentional or if CSFML just never supported them.

ChrisThrasher: Oh some of those weren't missing before. By casting the event type from the SFML type to the CSFML event type those events with no data members were being handled. Okay I see now. I'll add them.

src/SFML/Audio/SoundStream.cpp
ChrisThrasher: Why should `userData` remain the last parameter?

ChrisThrasher: Sounds good. I'll make that change.

eXpl0it3r (approved):

PR #249, opened 2024-05-18
Automatically Create a CSFML NuGet Package

eXpl0it3r:
Creating new releases is tedious work, as you need to build for multiple platforms, collect all the builds and then package it up.

With this new GitHub Action, it's all automated. It supports the following platforms:

- Windows x86
- Windows x64
- Linux x64
- Linux MUSL x64 (via Docker)
- Linux ARM (via QEMU)
- Linux ARM64 (via QEMU)
- Fedora x64 (via Docker
- macOS x64
- macOS ARM64

This PR also includes a commit to enable module selection for CSFML. This helps with build times of the NuGet workflow, as we don't need to build the network module, neither for SFML, nor CSFML.

I've left out automatic publishing to NuGet (for now), as this is a relatively simple step to do once you have the NuGet package and it's hard to tests, as you only want to upload the package for a new release.

To prevent unnecessary wait times (10+min) for our own braches, I've limited the NuGet CI execution to the `2.6.x` and `master` branch.


.github/workflows/nuget.yml
```diff
@@ -0,0 +1,252 @@
...
+              uses: actions/upload-artifact@v4
+              with:
+                  name: CSFML-win-x86-x64
+                  path: CSFML/tools/nuget/CSFML/runtimes
+
+    linux-x64:
+        name: Linux x64
+        runs-on: ubuntu-22.04
+        container:
+            image: ubuntu:22.04
```
kimci86, line 50: Isn't this redundant?

src/SFML/CMakeLists.txt
```diff
@@ -21,14 +21,37 @@ if(CSFML_LINK_SFML_STATICALLY)
...
+if(CSFML_BUILD_AUDIO)
+    list(PREPEND SFML_MODULES "audio")
+endif()
+if(CSFML_BUILD_GRAPHICS)
+    list(PREPEND SFML_MODULES "graphics")
+endif()
+if(CSFML_BUILD_NETWORK)
+    list(PREPEND SFML_MODULES "network")
+endif()
+find_package(SFML 2.6 COMPONENTS ${SFML_MODULES} REQUIRED)
```
kimci86, line 34: Missing instructions for the Window module.

src/SFML/CMakeLists.txt
```diff
@@ -21,14 +21,37 @@ if(CSFML_LINK_SFML_STATICALLY)
...
 
 # add the modules subdirectories
 add_subdirectory(System)
-add_subdirectory(Window)
-add_subdirectory(Network)
-add_subdirectory(Graphics)
-add_subdirectory(Audio)
+
+if(CSFML_BUILD_WINDOW OR SFML_BUILD_GRAPHICS)
```
kimci86, line 39: Typo
```suggestion
if(CSFML_BUILD_WINDOW OR CSFML_BUILD_GRAPHICS)
```

.github/workflows/nuget.yml
eXpl0it3r: I guess so. This comes from the time, I tried to run everything through Docker.

eXpl0it3r: Hmm, I'll leave it, so I don't have to install the dependencies natively, but can just keep it in the shell script.

src/SFML/CMakeLists.txt
eXpl0it3r: See #250

src/SFML/CMakeLists.txt
eXpl0it3r: See #250

PR #252, opened 2024-05-19
Create Release Packages Automatically

eXpl0it3r:
This is another step in automating releases. This prepares packages as we've been releasing as pre-built binaries for Windows and macOS. For Linux it's still advised to use official packages or build from source due to glibc and other compatibility issues.

For macOS was unable build for older macOS SDK version as we used to. With Apple not providing any reasonable support in this area and them heavily pushing users to new OS versions, I don't necessarily see this in our responsibility.

In case anyone wants to try. I got stuck on the compiler not finding the `<ostream>` header 🤷 

```
    macos:
        name: macOS
        runs-on: macos-14

        steps:
            - name: Cache macOS SDK 10.15
              id: macosx-10-15-sdk
              uses: actions/cache@v4
              with:
                  path: ${{ github.workspace }}/MacOSX10.15.sdk
                  key: macosx-10-15-sdk

            - name: Download macOS SDK 10.15
              if: steps.macosx-10-15-sdk.outputs.cache-hit != 'true'
              run: |
                curl -Lo MacOSX10.15.sdk.tar.xz https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX10.15.sdk.tar.xz
                tar -xvf MacOSX10.15.sdk.tar.xz

            - name: Install macOS SDK 10.15
              run: |
                  sudo mv MacOSX10.15.sdk /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/

            - name: Checkout SFML
              uses: actions/checkout@v4
              with:
                  repository: SFML/SFML
                  ref: 2.6.1
                  path: SFML

            - name: Checkout CSFML
              uses: actions/checkout@v4
              with:
                  path: CSFML

            - name: Build SFML x64
              run: |
                  cmake -S SFML -B SFML-build-x64 -G "Unix Makefiles" \
                        -DCMAKE_C_COMPILER="/usr/bin/clang" \
                        -DCMAKE_CXX_COMPILER="/usr/bin/clang++" \
                        -DCMAKE_C_FLAGS="-stdlib=libc++" \
                        -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
                        -DBUILD_SHARED_LIBS=TRUE \
                        -DCMAKE_BUILD_TYPE=Release \
                        -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/SFML-x64/Library/Frameworks \
[...]

PR #253, opened 2024-05-19
Update CSFML Version to 2.6.1

eXpl0it3r:
This is a new release of CSFML. With the automation added (#249 & #252), it should be quite easy to create a new release.

This is also quite important to bring missing features for SFML.Net, so we release a 2.6.0 version.


ChrisThrasher (approved):
🥳

PR #254, opened 2024-05-19
Temporarily lock SFML revision used in CI

ChrisThrasher:
The SFML 3 API is changing faster than I care to update in CSFML. *I can't stand whatever guy keeps breaking the SFML API*. Anyways, for now let's lock CSFML to a particular revision of SFML until the SFML 3 API changes slow down and we can go back to tracking the master branch.


ChrisThrasher:
Superceded by #255

PR #255, opened 2024-05-19
Update to latest SFML 3 API

ChrisThrasher:
Supersedes #254

### DIFF
--- a/src/SFML/Audio/SoundBuffer.cpp
+++ b/src/SFML/Audio/SoundBuffer.cpp
@@ -34,30 +34,22 @@
 ////////////////////////////////////////////////////////////
 sfSoundBuffer* sfSoundBuffer_createFromFile(const char* filename)
 {
-    sfSoundBuffer* buffer = new sfSoundBuffer;
+    auto soundBuffer = sf::SoundBuffer::loadFromFile(filename);
+    if (!soundBuffer)
+        return nullptr;
 
-    if (!buffer->This.loadFromFile(filename))
-    {
-        delete buffer;
-        buffer = nullptr;
-    }
-
-    return buffer;
+    return new sfSoundBuffer{std::move(*soundBuffer)};
 }
 
 
 ////////////////////////////////////////////////////////////
 sfSoundBuffer* sfSoundBuffer_createFromMemory(const void* data, size_t sizeInBytes)
 {
-    sfSoundBuffer* buffer = new sfSoundBuffer;
+    auto soundBuffer = sf::SoundBuffer::loadFromMemory(data, sizeInBytes);
+    if (!soundBuffer)
+        return nullptr;
 
-    if (!buffer->This.loadFromMemory(data, sizeInBytes))
-    {
-        delete buffer;
-        buffer = nullptr;
-    }
-
-    return buffer;
+    return new sfSoundBuffer{std::move(*soundBuffer)};
 }
 
 
@@ -67,15 +59,12 @@ sfSoundBuffer* sfSoundBuffer_createFromStream(sfInputStream* stream)
 {
     CSFML_CHECK_RETURN(stream, nullptr);
 
-    sfSoundBuffer* buffer = new sfSoundBuffer;
     CallbackStream sfmlStream(stream);
-    if (!buffer->This.loadFromStream(sfmlStream))
-    {
-        delete buffer;
-        buffer = nullptr;
-    }
+    auto soundBuffer = sf::SoundBuffer::loadFromStream(sfmlStream);
+    if (!soundBuffer)
+        return nullptr;
 
-    return buffer;
+    return new sfSoundBuffer{std::move(*soundBuffer)};
 }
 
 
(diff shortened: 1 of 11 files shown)


src/SFML/Audio/SoundBufferRecorder.cpp
```diff
@@ -70,9 +70,10 @@ const sfSoundBuffer* sfSoundBufferRecorder_getBuffer(const sfSoundBufferRecorder
 {
     CSFML_CHECK_RETURN(soundBufferRecorder, nullptr);
 
-    soundBufferRecorder->SoundBuffer.This = soundBufferRecorder->This.getBuffer();
+    soundBufferRecorder->SoundBuffer->This = soundBufferRecorder->This.getBuffer();
 
-    return &soundBufferRecorder->SoundBuffer;
+    assert(soundBufferRecorder->SoundBuffer && "sfSoundBufferRecorder_getBuffer() Cannot access null sound buffer");
```
eXpl0it3r, line 75: Can we, should do assert? It's the first of its kind (would also require an header include).

ChrisThrasher: I'm not 100% what I wrote is correct. I need to look at it some more.

src/SFML/Audio/SoundBufferRecorder.cpp
```diff
@@ -70,9 +70,10 @@ const sfSoundBuffer* sfSoundBufferRecorder_getBuffer(const sfSoundBufferRecorder
 {
     CSFML_CHECK_RETURN(soundBufferRecorder, nullptr);
 
-    soundBufferRecorder->SoundBuffer.This = soundBufferRecorder->This.getBuffer();
+    soundBufferRecorder->SoundBuffer->This = soundBufferRecorder->This.getBuffer();
```
kimci86, line 73: You need to fill the optional at some point.
```suggestion
    soundBufferRecorder->SoundBuffer = sfSoundBuffer{soundBufferRecorder->This.getBuffer()};
```
(untested)

src/SFML/Audio/SoundBufferRecorder.cpp
ChrisThrasher: I changed this to what Kimci recommended and think it's correct now.

src/SFML/Audio/SoundBuffer.cpp
```diff
@@ -34,30 +34,22 @@
 ////////////////////////////////////////////////////////////
 sfSoundBuffer* sfSoundBuffer_createFromFile(const char* filename)
 {
-    sfSoundBuffer* buffer = new sfSoundBuffer;
+    const auto soundBuffer = sf::SoundBuffer::loadFromFile(filename);
```
kimci86, line 37: This needs to be mutable to allow being moved from. (assuming SoundBuffer eventually implements move semantics)

ChrisThrasher: Sound buffers are already moveable so removing this `const` will enable a move. Thanks. I think clang-tidy will flag this but we don't have clang-tidy set up on CSFML yet.

src/SFML/Graphics/Font.cpp
```diff
@@ -64,15 +58,12 @@ sfFont* sfFont_createFromStream(sfInputStream* stream)
...
-        delete font;
-        font = nullptr;
-    }
+    CallbackStream sfmlStream(stream);
+    auto font = sf::Font::loadFromStream(sfmlStream);
+    if (!font)
+        return nullptr;
 
-    return font;
+    return new sfFont{std::move(*font), {}, sfmlStream};
```
kimci86, line 66: This is not correct: `sfmlStream` object passed by reference to `loadFromStream` must remain valid while the `sf::Font` instance is alive, but it is going out of scope.
Not sure what is best to solve this. An idea is having an `std::unique_ptr<CallbackStream>` in `sfFont`.

src/SFML/Audio/SoundBuffer.cpp
kimci86: To clarify my point: yes `sf::SoundBuffer` is moveable, but only as a consequence of being copyable. There is no specific move constructor/assignment operator that transfers ownership yet.

ChrisThrasher: https://github.com/SFML/SFML/blob/fdcd899d2b027bf61d11e69e44e55e6727ac6205/test/Audio/SoundBuffer.test.cpp#L20-L23

`sf::SoundBuffer` is in fact move constructible and move assignable.

src/SFML/Graphics/Font.cpp
ChrisThrasher: Looking at this function before and after, I'm not sure how anything about ownership changes. If you're correct then that would imply the state of this function on `master` also exhibits this issue.

kimci86: No, this is different. On master, the stream passed to `loadFromFile` which is `font->Stream` remains valid. It is owned by the `sfFont` instance.

ChrisThrasher: I'm confused. `sfmlStream` is used to construct the `sfFont` instance so the `sfFont` still owns it, does it not?

kimci86: `sfFont` is constructed with a copy of `sfmlStream`. The original stack variable `sfmlStream`, to which `sf::Font` internally has a pointer to and expects to remain valid, will become invalid at the end of the function.

ChrisThrasher: I think I'm following now. Good catch. Off the top of my head I can't picture a good way to fix this. Do you have any ideas?

kimci86: This discussion makes me realize `sfFont_copy` was probably wrong: making a copy of the `CallbackStream` instance while the `sf::Font` copy will still internally point to the original instance.

Anyways, I think this would be a correct solution: https://github.com/SFML/CSFML/compare/sfml3_updates...kimci86:CSFML:sfml3_updates. Not sure if there is some smarter solution.

ChrisThrasher: This makes sense to me. I think it's a correct solution.

ChrisThrasher:
Rebased on `master` and fixed even more API changes with SFML 3.

eXpl0it3r (approved):

PR #258, opened 2024-06-01
Documentation: Fix last value of index range in FTP listing response

mgrojo:
- Wording and notation consistent with `sf*Shape_{g,s}etPoint()`
- Related case fix in `sfConvexShape_setPoint`

### DIFF
--- a/include/CSFML/Graphics/ConvexShape.h
+++ b/include/CSFML/Graphics/ConvexShape.h
@@ -390,7 +390,7 @@ CSFML_GRAPHICS_API void sfConvexShape_setPointCount(sfConvexShape* shape, size_t
 /// of the valid range.
 ///
 /// \param shape Shape object
-/// \param index Index of the point to change, in range [0 .. GetPointCount() - 1]
+/// \param index Index of the point to change, in range [0 .. getPointCount() - 1]
 /// \param point New point
 ///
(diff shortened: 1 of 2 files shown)


include/SFML/Network/Ftp.h
```diff
@@ -168,7 +168,7 @@ CSFML_NETWORK_API size_t sfFtpListingResponse_getCount(const sfFtpListingRespons
 /// \brief Return a directory/file name contained in a FTP listing response
 ///
 /// \param ftpListingResponse Ftp listing response
-/// \param index              Index of the name to get (in range [0 .. getCount])
+/// \param index              Index of the name to get, in range [0 .. getCount() - 1]
```
eXpl0it3r, line 171: Seems like you're still missing @ChrisThrasher change request

```suggestion
/// \param index              Index of the name to get, in range [0 .. getCount())
```

mgrojo: Ok, ok, that was intentional, but I'll accommodate to your liking.

`sf*Shape_{g,s}etPoint()` will still be using another notation, which is more explicit and clear, in my opinion.

ChrisThrasher:
I reverted the PR back to its original state since that was better than what I was recommending. Thanks for the fix!

PR #264, opened 2024-08-09
Update to latest SFML 3 API

ChrisThrasher:


eXpl0it3r (approved):
Something to add to the API checking list is, why we didn't have to change anything for the `create` / `resize` API change, i.e. there's probably some missing API there. 

The NuGet build will eventually fail (again), but I'll have to fix that separately.

PR #265, opened 2024-08-20
Revert "Temporarily lock SFML revision used in CI"

ChrisThrasher:
This reverts commit 36a6e7c5a617b3f6cdb6bac9a9613e516ea6ba6e.

Now that SFML 3's API is mostly stable we no longer need to lock down the specific commit hash.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: SFML/SFML
-        ref: 8096ba24fccb6b76a7d9e8cc71a3afb7b02e517c
+        ref: master
         path: SFML
 
     - name: Configure SFML CMake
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
https://github.com/SFML/CSFML/actions/runs/10464568981/job/28978624152?pr=265#step:9:189

This error has happened before and it's been fixed by rerunning the job. I now see it's a race condition between compiling Catch2 and copying its DLLs. I fixed it in another commit in this PR.

kimci86:
I believe the build error is caused by several copies of the same file running at the same time.

One way to solve it as this CMake [documentation example](https://cmake.org/cmake/help/latest/command/add_custom_command.html#example-generating-files-for-multiple-targets) may suggest is to do the copy in a single custom target.
I tried to implement this here: https://github.com/kimci86/CSFML/tree/bugfix/test_dll_copy_concurrency
I am not sure how to get rid of the stamp file I am using in that branch.

PR #270, opened 2024-08-24
CMake: Fix installing HTML Help with newer Doxygen version

Biswa96:
HTML Help generation was disabled for doxygen version greater than 1.10.0 in df75efa78d41e39427168355659e48f8d2e6e7d8 commit.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -48,7 +48,7 @@ add_custom_target(doc ALL
 install(DIRECTORY ${DOXYGEN_OUTPUT_DIR}/html
         DESTINATION ${INSTALL_MISC_DIR}/doc
         COMPONENT doc)
-if(DOXYGEN_HHC_PROGRAM)
+if(DOXYGEN_GENERATE_HTMLHELP)
     install(FILES ${DOXYGEN_OUTPUT_DIR}/CSFML.chm
             DESTINATION ${INSTALL_MISC_DIR}/doc
             COMPONENT doc)


eXpl0it3r:
What does this change do exactly?

The disabling was done, because doxygen caused problems with certain parameters when having HTML Help and HTML enabled at the same time.

Biswa96:
> What does this change do exactly?

With newer doxygen, the HTML Help file is not generate and failed to install with the following error.

```
CMake Error at build/doc/cmake_install.cmake:45 (file):
  file INSTALL cannot find
  "csfml/src/build/doc/CSFML.chm": File
  exists.
Call Stack (most recent call first):
  build/cmake_install.cmake:55 (include)
```

ChrisThrasher (approved):
I cannot reproduce this on macOS but I believe that's because `DOXYGEN_HHC_PROGRAM` is Windows-specific.

eXpl0it3r:
Will need to be fixed in SFML as well:
image

eXpl0it3r:
As `DOXYGEN_GENERATE_HTMLHELP` is only set to true when `DOXYGEN_HHC_PROGRAM` is true, it's redudant to check both, thus I changed it.

Since this was already done on 2.6.1, I've rebased the branch onto `2.6.x` and changed the PR target.

eXpl0it3r:
Well the mentioned error is actually about not having the doc target built first, additionally, that was for SFML not CSFML, but I guess that shouldn't matter as much.

image

Biswa96:
It seems that sfml also need same changes. I get the following error.

```

CMake Error at bin/doc/cmake_install.cmake:41 (file):
  file INSTALL cannot find "SFML/bin/doc/SFML.tag": No error.
Call Stack (most recent call first):
  bin/cmake_install.cmake:280 (include)

CMake Error at bin/doc/cmake_install.cmake:45 (file):
  file INSTALL cannot find "SFML/bin/doc/sfml.chm": File exists.
Call Stack (most recent call first):
  bin/cmake_install.cmake:280 (include)
```

CMake shows the following output after finding doxygen.

```
-- Found Doxygen: C:/msys64/ucrt64/bin/doxygen.exe (found version "1.12.0") found components: doxygen dot
Due to conflicts with the HTML output settings in Doxygen 1.12.0, the HTML Help generation will be disabled
```

Please let me know if I am missing something.

eXpl0it3r:
Thank you for the reminder!

I've created two pull requests, for SFML 2.6.3 and 3.0.0

As we're currently focusing on SFML 3, the 2.6.3 merge has to wait a bit.

PR #273, opened 2024-08-25
Added functions to convert between corresponding SFML and CSFML types

ChrisThrasher:
This removes a lot of code that was doing the same thing over and over again. I chose the name `convert` since it's nice and short. There are similar functions named `convertEvent`, `convertTransform`, and `convertRenderStates` that do similar things for other types. I'd like to rename those to `convert` if the team is onboard with that but it's also fine to leave them as-is.


src/CSFML/Graphics/CircleShape.cpp
```diff
@@ -243,13 +228,7 @@ sfColor sfCircleShape_getFillColor(const sfCircleShape* shape)
...
     CSFML_CHECK_RETURN(shape, color);
 
-    sf::Color sfmlColor = shape->This.getFillColor();
-    color.r = sfmlColor.r;
-    color.g = sfmlColor.g;
-    color.b = sfmlColor.b;
-    color.a = sfmlColor.a;
-
-    return color;
+    return convert(shape->This.getFillColor());
```
eXpl0it3r, line 231: I find `convert` on its own problematic, as seen here, you'll need to know that we're not turning the color into something different. I'd personally never use such a generic name for a free function.

eXpl0it3r (approved):

PR #283, opened 2024-09-01
Remove unnecessary `extern` from C function declarations

ChrisThrasher:


ChrisThrasher:
It's true that `extern` on C function declarations is unnecessary. However, these macros are also used for global variables and I think we do need to mark those as `extern` which currently is not happening.

PR #284, opened 2024-09-01
Add clang-tidy `modernize-` checks

ChrisThrasher:


ChrisThrasher:
Certain checks like `modernize-use-using` have to be disabled because our C headers use `typedef`s. I'm not yet sure of a good way to disable this check for our .h files while keeping it enabled for everything else.

PR #295, opened 2024-09-03
Add clang-tidy `cppcoreguidelines-*` checks

ChrisThrasher:


src/CSFML/Graphics/TextureStruct.hpp
```diff
@@ -35,22 +35,14 @@
...
     {
-        This        = texture;
-        OwnInstance = false;
     }
 
-    sfTexture(const sfTexture& texture)
+    sfTexture(const sfTexture& texture) : This(texture.This ? new sf::Texture(*texture.This) : nullptr)
     {
-        This        = texture.This ? new sf::Texture(*texture.This) : nullptr;
-        OwnInstance = true;
```
eXpl0it3r, line 53: I don't know what `OwnInstance` is used for, but this changes semantics, as the `sf::Texture*` constructor overload will set `OwnInstance` to `false`, while this constructor is supposed to set it to `true`.

ChrisThrasher: `OwnInstance` determines whether `delete` is called on `This`. I changed the default value of `OwnInstance` to `true` so its value in this constructor shouldn't be different than before.

eXpl0it3r: Ah right, I misread that

eXpl0it3r (approved):

PR #302, opened 2024-09-03
Clarify instructions for building CSFML

ChrisThrasher:
Building CSFML isn't particularly complicated and for the most part requires the same CMake knowledge required to build any library but I still thought it was reasonable to include a few bullet points to help demystify the process.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -36,7 +36,19 @@ Of course, you can also find the CSFML API documentation in the SDK.
 
 ## Building
 
-Set `SFML_ROOT` to a directory containing the installation of SFML.
+Building CSFML follows typical CMake conventions.
+Here's how to get started.
+
+1. Install SFML.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -36,7 +36,19 @@ Of course, you can also find the CSFML API documentation in the SDK.
...
 ## Building
 
-Set `SFML_ROOT` to a directory containing the installation of SFML.
+Building CSFML follows typical CMake conventions.
+Here's how to get started.
+
+1. Install SFML.
+This may be done in a variety of ways including using a system package manager or building and installing SFML from source.
+The major version of SFML must match the major version of CSFML.
+1. Configure CSFML.
```
eXpl0it3r, line 45: Since Markdown files can/should also be readable in plaintext, I'd personally stick with continued numbering (1., 2., 3., ...)

eXpl0it3r (approved):

PR #327, opened 2024-09-12
Improve const correctness of `_destroy` functions

ChrisThrasher:
It's perfectly valid to call `delete` on a pointer-to-const. However, all CSFML `_destroy` functions use a pointer-to-non-const. One downside of this approach is that you cannot create an object and bind it to a `const T*`. Doing so will make it impossible to call that object's `_delete` function.

### DIFF
--- a/examples/example.c
+++ b/examples/example.c
@@ -12,7 +12,7 @@ int main(void)
         return EXIT_FAILURE;
 
     // Load a sprite to display
-    sfTexture* texture = sfTexture_createFromFile("sfml_logo.png", NULL);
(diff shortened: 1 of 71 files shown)


examples/example.c
```diff
@@ -23,7 +23,7 @@ int main(void)
     sfSprite_setPosition(sprite, spritePosition);
 
     // Create a graphical text to display
-    sfFont* font = sfFont_createFromFile("tuffy.ttf");
```
ChrisThrasher, line 26: Adding `const` here causes a compiler error on CSFML `master` as it stands today. There's no good reason to prevent this.

ZXShady: yea we don't do the same mistake as C free.

PR #347, opened 2024-09-20
Use inheritance to remove `This` data member

### LINKED ISSUE #222: Use inheritance to implement wrapper struct types
We can save code and remove the ugly `This` parameter if we instead inherit from `sf::` types.

Here's an example: https://github.com/SFML/CSFML/compare/master...simplify_sf_buffer

ChrisThrasher:
Closes #222

It's reasonable to say that, for example, `sfImage` _is_ an `sf::Image`. This "is-a" relationship is what inheritance does a very good job modeling. By using inheritance to implement CSFML types in terms of their SFML counterpart, we get code that is simpler and easier to read.


ZXShady (changes requested):
remove unnecessary ternary and static cast

src/CSFML/Graphics/Shader.cpp
```diff
@@ -243,8 +243,7 @@ void sfShader_setColorUniform(sfShader* shader, const char* name, sfColor color)
 void sfShader_setIntUniform(sfShader* shader, const char* name, int x)
 {
     assert(shader);
-    assert(name);
```
ZXShady, line 246: is there a reason for removing this assert? might have been accidental

src/CSFML/Graphics/Shader.cpp
```diff
@@ -419,22 +418,22 @@ void sfShader_setTextureParameter(sfShader* shader, const char* name, const sfTe
...
-    return shader->This.getNativeHandle();
+    return shader->getNativeHandle();
 }
 
 
 ////////////////////////////////////////////////////////////
 void sfShader_bind(const sfShader* shader)
 {
-    sf::Shader::bind(shader ? &shader->This : nullptr);
+    sf::Shader::bind(shader ? static_cast<const sf::Shader*>(shader) : nullptr);
```
ZXShady, line 436: the static cast is unnecessary and the ternary is also unnecessary

the implicit Derived to Base conversion is enough.

ZXShady (changes requested):
readd assert and remove unnecessary static_cast and ternary

eXpl0it3r:
Could this cause issues from pure C side or a binding using CSFML? We're only returning pointers and have typedefs, but now the fundamental type is no longer just a simple struct 🤔

ZXShady:
the type was never a simple struct to begin with the sf Classes had non standard layout types inside it. and the typedef is to make the pointer opaque.

ChrisThrasher:
> Could this cause issues from pure C side or a binding using CSFML? We're only returning pointers and have typedefs, but now the fundamental type is no longer just a simple struct 🤔

Notice how this PR only touches the `src/` directory. Users have no way to observe structs that are defined in `src/`. They can only see the definition of structs defined in `include/`. This is the reason we use opaque pointers in the interface. It gives us full freedom to implement these structs however we like.

eXpl0it3r:
I get that, I was just wondering if there's any limit to that. We now had opaque pointers to structs, but now it's a derived type. Can this have any side effect from a C/binding point of view?

ChrisThrasher:
No, there’s no way for users to notice anything has changed.

ZXShady (approved):

PR #358, opened 2024-09-27
use aggregate initialization to initialize

ZXShady:
I personally think it is clearer than the variable + member assignments as the layout of sfColor is guranteed and known and this removes some intermediary variables.

### DIFF
--- a/src/CSFML/Graphics/Transform.cpp
+++ b/src/CSFML/Graphics/Transform.cpp
@@ -48,8 +48,13 @@ const sfTransform sfTransform_Identity = {
 ////////////////////////////////////////////////////////////
 sfTransform sfTransform_fromMatrix(float a00, float a01, float a02, float a10, float a11, float a12, float a20, float a21, float a22)
 {
-    sfTransform transform = {a00, a01, a02, a10, a11, a12, a20, a21, a22};
-    return transform;
+    // clang-format off
+    return {
+        a00, a01, a02,
+        a10, a11, a12,
+        a20, a21, a22,
+    };
+    // clang-format on
 }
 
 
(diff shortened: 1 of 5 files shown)


src/CSFML/Graphics/Transform.cpp
```diff
@@ -48,8 +48,13 @@ const sfTransform sfTransform_Identity = {
 ////////////////////////////////////////////////////////////
 sfTransform sfTransform_fromMatrix(float a00, float a01, float a02, float a10, float a11, float a12, float a20, float a21, float a22)
 {
-    sfTransform transform = {a00, a01, a02, a10, a11, a12, a20, a21, a22};
-    return transform;
+    // clang-format off
+    return {
+        a00,a01,a02,
+        a10,a11,a12,
+        a20,a21,a22,
```
ChrisThrasher, line 55:
```suggestion
        a00, a01, a02,
        a10, a11, a12,
        a20, a21, a22,
```

ChrisThrasher (changes requested):

ChrisThrasher (approved):

src/CSFML/System/Time.cpp
```diff
@@ -56,25 +56,19 @@ int64_t sfTime_asMicroseconds(sfTime time)
...
 }
 
 
 ////////////////////////////////////////////////////////////
 sfTime sfMilliseconds(int32_t amount)
 {
-    sfTime time;
-    time.microseconds = static_cast<int64_t>(amount) * 1000;
-    return time;
+    return {static_cast<int64_t>(amount * 1000)};
```
kimci86, line 66: For the record, bug here got fixed in #360

PR #360, opened 2024-09-28
cast to larger type before doing arithmetic to avoid signed integer o…

ZXShady:
…verflow

https://godbolt.org/z/soTad4zEq

this godbolt links shows that it is UB to multiply then cast if the amount is large enough

### DIFF
--- a/src/CSFML/System/Time.cpp
+++ b/src/CSFML/System/Time.cpp
@@ -63,7 +63,7 @@ sfTime sfSeconds(float amount)
 ////////////////////////////////////////////////////////////
 sfTime sfMilliseconds(int32_t amount)
 {
-    return {static_cast<int64_t>(amount * 1000)};
+    return {int64_t{amount} * 1000};
 }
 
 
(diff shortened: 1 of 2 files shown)


src/CSFML/System/Time.cpp
```diff
@@ -56,14 +56,14 @@ int64_t sfTime_asMicroseconds(sfTime time)
 ////////////////////////////////////////////////////////////
 sfTime sfSeconds(float amount)
 {
-    return {static_cast<int64_t>(amount * 1000000)};
+    return {static_cast<int64_t>(amount) * 1000000};
```
Marioalexsan, line 59: If I pass `1.52f`, it's going to be truncated to `1` with such a cast.

ZXShady: haha I didn't realize that also float is definitely big enough to hold all values of uint64_t I should only change for the other function, thanks for noticing

src/CSFML/System/Time.cpp
```diff
@@ -63,7 +63,7 @@ sfTime sfSeconds(float amount)
 ////////////////////////////////////////////////////////////
 sfTime sfMilliseconds(int32_t amount)
 {
-    return {static_cast<int64_t>(amount * 1000)};
+    return {static_cast<int64_t>(amount) * 1000};
```
ChrisThrasher, line 66:
```suggestion
    return {int64_t{amount} * 1000};
```
Because this is simply a promotion, we can avoid the static cast.

ChrisThrasher (changes requested):

ChrisThrasher (approved):

ChrisThrasher (approved):

PR #361, opened 2024-09-28
check malloc return value for NULL and update docs

ZXShady:
`sfFtpDirectoryResponse_getDirectory` can fail and return NULL becuase 'strdup' returns null on failure and it didn't mention it in the docs but `sfFtpDirectoryResponse_getDirectoryUnicode` didn't and just had UB due not checking return value of malloc.

typing `NULL` instead of `nullptr` in the docs made me age 30 years btw

### DIFF
--- a/src/CSFML/Network/Ftp.cpp
+++ b/src/CSFML/Network/Ftp.cpp
@@ -51,6 +51,8 @@ static_assert(alignof(sfChar32) == alignof(char32_t));
 {
     const std::size_t byteCount = sizeof(sfChar32) * str.getSize();
     auto*             utf32     = static_cast<sfChar32*>(std::malloc(byteCount + sizeof(sfChar32)));
+    if (!utf32)
+        return nullptr;
     std::memcpy(utf32, str.getData(), byteCount);
     utf32[str.getSize()] = 0;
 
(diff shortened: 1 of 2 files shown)


src/CSFML/Network/Ftp.cpp
```diff
@@ -49,11 +49,11 @@ static_assert(alignof(sfChar32) == alignof(char32_t));
...
-    const std::size_t byteCount = sizeof(sfChar32) * str.getSize();
-    auto*             utf32     = static_cast<sfChar32*>(std::malloc(byteCount + sizeof(sfChar32)));
+    const std::size_t byteCount = sizeof(sfChar32) * (str.getSize() + 1);
+    auto*             utf32     = static_cast<sfChar32*>(std::malloc(byteCount));
+    if (!utf32)
+        return nullptr;
     std::memcpy(utf32, str.getData(), byteCount);
-    utf32[str.getSize()] = 0;
-
     return utf32;
```
ChrisThrasher, line 57: Can you talk me through why `byteCount` changed and why `utf32[str.getSize()] = 0;` was removed? The commit message only talks about checking a return value so I was not expecting to see deeper changes to the implementation.

ZXShady: I didn't like the repeating of the typename sfChar32 twice (once in malloc call and once in byteCount) and I used the fact that C++11 strings are now always null terminated in the memcpy call and can now remove explicit setting of null terminator.

should have mentioned it in the pr message.

ChrisThrasher (approved):

PR #362, opened 2024-09-28
make static variables const via lamdbas

ZXShady:
this pr also makes `sfSoundRecorder_getAvailableDevices` return a const pointer to a const char pointer so you can no longer modify where the pointers of the strings are pointing into becuase it doesn't make sense I will remove it if it is unwanted

### DIFF
--- a/src/CSFML/Audio/SoundRecorder.cpp
+++ b/src/CSFML/Audio/SoundRecorder.cpp
@@ -81,18 +81,17 @@ bool sfSoundRecorder_isAvailable()
 
 
 ////////////////////////////////////////////////////////////
-const char** sfSoundRecorder_getAvailableDevices(size_t* count)
+const char* const* sfSoundRecorder_getAvailableDevices(size_t* count)
 {
-    static std::vector<std::string> stringDevices = sf::SoundRecorder::getAvailableDevices();
-    static std::vector<const char*> cstringDevices;
-
-    if (cstringDevices.empty() && !stringDevices.empty())
+    static const auto cstringDevices = []
     {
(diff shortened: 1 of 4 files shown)


src/CSFML/Window/VideoMode.cpp
```diff
@@ -41,14 +41,13 @@ sfVideoMode sfVideoMode_getDesktopMode()
 ////////////////////////////////////////////////////////////
 const sfVideoMode* sfVideoMode_getFullscreenModes(size_t* count)
 {
-    static std::vector<sfVideoMode> modes;
-
-    // Populate the array on first call
-    if (modes.empty())
+    static const auto modes = []()
```
ChrisThrasher, line 44:
```suggestion
    static const auto modes = []
```

src/CSFML/Audio/SoundRecorder.cpp
```diff
@@ -81,18 +81,17 @@ bool sfSoundRecorder_isAvailable()
...
 
 ////////////////////////////////////////////////////////////
-const char** sfSoundRecorder_getAvailableDevices(size_t* count)
+const char* const* sfSoundRecorder_getAvailableDevices(size_t* count)
 {
-    static std::vector<std::string> stringDevices = sf::SoundRecorder::getAvailableDevices();
-    static std::vector<const char*> cstringDevices;
-
-    if (cstringDevices.empty() && !stringDevices.empty())
+    static const auto cstringDevices = []()
```
ChrisThrasher, line 86:
```suggestion
    static const auto cstringDevices = []
```

ChrisThrasher (changes requested):

ChrisThrasher (approved):

PR #366, opened 2024-10-02
Make enumeration counts constant expressions

ZXShady:
This change allows now having these variables as array sizes that are not Variable Length Arrays like in structs for example.

```cpp
struct {
   sfKeycode keys[sfKeyCount]; // didn't compile since it was not constant expression
} global_table;

int main(void) {
   return 0;
}
```

To note however that this changes the type from unsigned int to an implementation defined type which usually `int` which is signed and not unsigned. and also disallows taking the address of the constant since it is mo longer a variable.

the other is using a macro which I don't like, but it can allow us to change the type to an unsigned type if we really wanted

```cpp
#define sfKeyCount ((unsigned int)sfKeyPause + 1)
```

but I prefer CSFML to be macro free if possible so I choose the enum approach.

### DIFF
--- a/test/Window/Keyboard.test.cpp
+++ b/test/Window/Keyboard.test.cpp
@@ -14,7 +14,7 @@ TEST_CASE("[Window] sfKeyboard")
         STATIC_CHECK(sfKeyC == static_cast<int>(sf::Keyboard::Key::C));
         STATIC_CHECK(sfKeyD == static_cast<int>(sf::Keyboard::Key::D));
         STATIC_CHECK(sfKeyPause == static_cast<int>(sf::Keyboard::Key::Pause));
-        CHECK(sfKeyCount == sf::Keyboard::KeyCount);
+        STATIC_CHECK(sfKeyCount == sf::Keyboard::KeyCount);
     }
 
     SECTION("sfScancode")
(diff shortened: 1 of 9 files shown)


eXpl0it3r:
Doesn't that still need to be exported?

ZXShady:
enums do not provide functionality I don't think they need to be exported and as I see no enums in CSFML are exported

ChrisThrasher:
For the sake of consistency could you make this change to all similar constants? I think it's just `sfMouseButtonCount`  and `sfSensorCount` that need this treatment.

ChrisThrasher (approved):

test/Window/Keyboard.test.cpp
```diff
@@ -14,7 +14,7 @@ TEST_CASE("[Window] sfKeyboard")
         STATIC_CHECK(sfKeyC == static_cast<int>(sf::Keyboard::Key::C));
         STATIC_CHECK(sfKeyD == static_cast<int>(sf::Keyboard::Key::D));
         STATIC_CHECK(sfKeyPause == static_cast<int>(sf::Keyboard::Key::Pause));
-        CHECK(sfKeyCount == sf::Keyboard::KeyCount);
+        STATIC_CHECK(sfKeyCount == sf::Keyboard::KeyCount);
```
ZXShady, line 17: Keyboard::KeyCount is unsigned int while sfKeyCount is signed wouldn't this cause warnings?

ChrisThrasher: If CI doesn't emit any warnings then I'm not worried.

PR #388, opened 2025-01-12
Update Network module code

Marioalexsan:
Should be the last changes needed to align CSFML and SFML, theoretically.

I changed the representation of `sfIpAddress_None` so that it is distinct from `sfIpAddress_Any` and actually represents an invalid / empty address, i.e. all bytes set to 0. Not 100% sure if this is meaningful or if it changes any behaviour in the code.


include/CSFML/Network/TcpListener.h
```diff
@@ -127,3 +142,12 @@ CSFML_NETWORK_API sfSocketStatus sfTcpListener_listen(sfTcpListener* listener, u
...
 CSFML_NETWORK_API sfSocketStatus sfTcpListener_accept(sfTcpListener* listener, sfTcpSocket** connected);
+
+////////////////////////////////////////////////////////////
+/// \brief Return the special value that tells the system
+///        to pick any available port
+///
+/// \return The value to use for any port
+///
+////////////////////////////////////////////////////////////
+CSFML_NETWORK_API unsigned short sfTcpListener_anyPort(void);
```
ChrisThrasher, line 153:
```suggestion
CSFML_NETWORK_API unsigned short sfTcpListener_anyPort(void);

```

src/CSFML/Network/TcpListener.cpp
```diff
@@ -102,3 +110,10 @@ sfSocketStatus sfTcpListener_accept(sfTcpListener* listener, sfTcpSocket** conne
 
     return status;
 }
+
+
+////////////////////////////////////////////////////////////
+unsigned short sfTcpListener_anyPort(void)
```
ChrisThrasher, line 116:
```suggestion
unsigned short sfTcpListener_anyPort()
```
`(void)` is required in the headers because the headers are C but we can omit that in our C++ implementation files.

src/CSFML/Network/UdpSocket.cpp
```diff
@@ -203,3 +203,10 @@ unsigned int sfUdpSocket_maxDatagramSize()
 {
     return sf::UdpSocket::MaxDatagramSize;
 }
+
+
+////////////////////////////////////////////////////////////
+unsigned short sfUdpSocket_anyPort(void)
```
ChrisThrasher, line 209:
```suggestion
unsigned short sfUdpSocket_anyPort()
```

src/CSFML/Network/IpAddress.cpp
```diff
@@ -48,7 +48,7 @@ namespace
 
 
 ////////////////////////////////////////////////////////////
-const sfIpAddress sfIpAddress_None = sfIpAddress_fromBytes(0, 0, 0, 0);
+const sfIpAddress sfIpAddress_None = {{0}};
```
ChrisThrasher, line 51: Is this not identical to what we previously had? I believe `sfIpAddress_fromBytes(0, 0, 0, 0)` should produce an array of all zeros.

ChrisThrasher (changes requested):

src/CSFML/Network/IpAddress.cpp
Marioalexsan: It would produce the "0.0.0.0" address, but this isn't what `sfIpAddress_None` should be. It should represent the same state as `std::nullopt` for `std::optional<sf::IpAddress>`, i.e. an invalid / missing IP address. Initializing the char array with zeroes should make it act as an empty string, which would make it distinct from `sfIpAddress_Any`.

ChrisThrasher (approved):

PR #390, opened 2025-01-15
Update Release Builds

eXpl0it3r:
Fixes the release build for SFML 3

Still some fine tuning is required:
- [x] Reenable branch filter
- [x] Correctly package CSFML on Windows - It was the weirdest PowerShell behavior/bug: https://stackoverflow.com/questions/79376139/why-does-move-item-into-a-new-directory-implode-the-first-directory

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,14 +24,14 @@ jobs:
               uses: actions/cache@v4
               with:
                   path: ${{ github.workspace }}/mingw32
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on: [workflow_dispatch]
```
ChrisThrasher, line 3: Why would we want to stop running CI for pushes and PRs?

ChrisThrasher (approved):

PR #393, opened 2025-02-16
Avoid doing a real name resolution in sfIpAddress_fromString test

jcowgill:
On some PCs, resolving `invalid address` will perform a real name resolution using the network. This is slows down the test and risks the test failing if the resolution succeeds.

Use `localhost` instead which shold return a known value, and add an invalid ip address test for good measure.


eXpl0it3r (approved):
I think that's a fair trade-off

ChrisThrasher:
Why not just use an even more nonsense name like a random 40 character string? We can certain solve the problem accidentally using a hostname present on the system. After we do that, are there still renaming issues we need to address?

eXpl0it3r:
The issue is the same as in the other PR. The goal isn't to make network calls during test execution.

PR #177, opened 2022-12-16
added functions for creating sRGB enabled textures from file / memory, etc.

EnokViking:
..and removed some unused functions, sfTexture_setSrgb and sfTexture_loadFromFile.

PR #408, opened 2025-08-31
Split texture's resize function into sRGB and non-sRGB

eXpl0it3r:
With #177 we've implemented `sfTexture_resize`, but the function takes a `bool` for switching between RGB and non-RGB, which is counter to the `sfTexture_createX` methods that each have their own sRGB versions.

There are two ways forward:

- Merge the `createX` and `createSrgbX` methods
- Split the `resize` method

I believe the reason for the separate functions are two fold:

- Adding functionality without API break
- C doesn't know of default parameter values, so method duplication is the answer

This PR chooses the method split, so one doesn't need to always provide the boolean.


ChrisThrasher (approved):

dogunbound:
Hmm, can't seem to find the approval button on GitHub mobile. This is my approval :)

Issue #409, opened 2025-09-02
NuGet DLLs depend on MSVC runtime libraries


JupiterRider:
Hello,

I noticed that the Windows DLLs in the new [CSFML NuGet packages](https://www.nuget.org/packages/CSFML/3.0.0-rc.2) depend on the MSVC runtime libraries (msvcp140.dll, vcruntime140.dll, etc.). CSFML 2.* didn't.

I'm afraid that this will lead to DllNotFoundExceptions and confusion for some users.

I used [python-pefile](https://github.com/erocarrera/pefile/) to list the dependencies: 

```python
>>> import pefile
>>> pe = pefile.PE("CSFML.3.0.0-rc.2/runtimes/win-x64/native/csfml-graphics.dll")
>>> for entry in pe.DIRECTORY_ENTRY_IMPORT:
...     print(entry.dll)
...     
b'sfml-graphics-3.dll'
b'sfml-window-3.dll'
b'sfml-system-3.dll'
b'MSVCP140.dll'
b'VCRUNTIME140.dll'
b'VCRUNTIME140_1.dll'
b'api-ms-win-crt-heap-l1-1-0.dll'
b'api-ms-win-crt-string-l1-1-0.dll'
b'api-ms-win-crt-runtime-l1-1-0.dll'
b'api-ms-win-crt-locale-l1-1-0.dll'
b'KERNEL32.dll'
>>> pe = pefile.PE("CSFML.2.6.0/runtimes/win-x64/native/csfml-graphics.dll")
>>> for entry in pe.DIRECTORY_ENTRY_IMPORT:
...     print(entry.dll)
...     
b'OPENGL32.dll'
b'WINMM.dll'
b'GDI32.dll'
b'KERNEL32.dll'
b'USER32.dll'
b'ADVAPI32.dll'
>>>```

eXpl0it3r:
Fixed with #412

PR #410, opened 2025-09-02
enable static build for windows

JupiterRider:
This should solve #409


JupiterRider:
I cam closing this in favor of #412

Issue #136, opened 2019-03-08
Joystick state does not update when polling events on Windows


caldur:
Hi,

The joystick state does not seem to get updated properly on Windows.

According to the tutorial, "Joystick states are automatically updated when you check for events”, which I assume means that they got updated when the window is polling events. Doc on `sfJoystick_update` as well as the official sfml joystick example also work as suggested.

However when testing the following code with CSFML 2.5, calls to `sfJoystick_isConnected` always return false, unless I manually update the joystick state before the calls.

`
sfRenderWindow* window = sfRenderWindow_create(mode, "joystick test", sfClose, NULL);

  printf("Before polling:\n");
  for (int i = 0; i < sfJoystickCount; ++i) {
    // sfJoystick_update();
    sfBool connected = sfJoystick_isConnected(i);
    printf("%d: %d\n", i, connected);
  }
  
  printf("Polling starts:\n");

  while(sfRenderWindow_isOpen(window)) {
    sfEvent evt;
    while(sfRenderWindow_pollEvent(window, &evt)) {
      //sfJoystick_update();
      if (evt.type == sfEvtClosed) {
        sfRenderWindow_close(window);
        break;
      }

      else if (evt.type == sfEvtJoystickMoved) {
        //sfJoystick_update();
        unsigned int id = evt.joystickMove.joystickId;
        printf("%d: %d\n", id, sfJoystick_isConnected(id));
      }
    }
  }
`
I am running building the above code with CSFML 2.5 64bit, mingw gcc 8.1 on windows 7. The joystick used is a compatible XBOX controller.
Similar code in c++ with SFML 2.5.1 runs without issue, however.
I also tested the code on mac and it worked as intended, so I'd suspect it's a windows only issue?

Any insights would be appreciated.

eXpl0it3r:
Thanks for reporting this issue!

Maybe it's not propagates to the correct C++ class/pointer. Requires further investigation.

andre-sampaio:
I have been having the same [problem](https://github.com/SFML/SFML.Net/issues/189) with SFML on .Net. After some investigation I realized that building SFML with the flag BUILD_SHARED_LIBS=1 and SFML_USE_STATIC_STD_LIBS=0 and then building CSML with the flag CSFML_LINK_SFML_STATICALLY=0 makes this sample work as expected (I have been using tools\nuget\build.win.ps1 to build). You can check that in my [fork](https://github.com/andre-sampaio/CSFML).

[...]

PR #412, opened 2025-09-03
Package runtime DLLs for NuGet

eXpl0it3r:
Include the CRT DLLs with the NuGet package fixing #409 

- We can't link the CRT statically, as this would require us to link SFML statically, which would again trigger #136 
- We need to ensure to only ship DLLs that are part of the redist to adhere to Microsoft's license agreement

@JupiterRider Do you have an easy way to test a NuGet package without Redist already installed?
If so, can you given the following NuGet package a try?

https://github.com/SFML/CSFML/actions/runs/17425612541/artifacts/3913669818

### DIFF
--- a/tools/nuget/build.win.ps1
+++ b/tools/nuget/build.win.ps1
@@ -13,6 +13,7 @@ if (-not $RID) {
     Write-Output "No RID specified, building all known RIDs"
     ./build.win.ps1 "win-x86"
     ./build.win.ps1 "win-x64"
+    ./build.win.ps1 "win-arm64"
     exit
 }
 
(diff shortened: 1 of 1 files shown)


JupiterRider:
I successfully tested the libraries in the Nuget package inside a blank x64 Windows 10 VM. 

I used my own application which loads csfml-graphics.dll via `LoadLibraryExA("assets/csfml-graphics.dll", NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS)`.

eXpl0it3r:
Awesome, thanks for testing! 🙂

PR #414, opened 2025-09-13
Mention third-party resource in readme.md

JupiterRider:
Sigmanificient created a CSFML API Doc and I think this could be helpful:
- Github Repo: https://github.com/Sigmanificient/csfml-handbook
- Hosted API Docs: https://csfml.1l.is/api-v3/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -26,6 +26,8 @@ You can get the latest official release on [SFML website](https://www.sfml-dev.o
 
 ## Learn
 
+### Offical resources
+
 There is no tutorial for CSFML, but since it's a binding you can use the C++ resources:
 
   * The [official tutorials](https://www.sfml-dev.org/tutorials/)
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
While I appreciate the effort in hosting this and can see that it's useful for others, I don't think we can/should link to things that are not under our control, as we can't guarantee the availability over the timespan that CSFML will be relevant.

Sigmanificient:
@eXpl0it3r I can grant you control over the repo if you are interested tho :)

eXpl0it3r:
Repo access is one thing, the more important is the hosting location and access. I'm weary to host something "official" at 1l.is

If you're interested in getting the CSFML documentation hosted on the official SFML website, I'm more than happy to review a PR against the [SFML-Website repository](https://github.com/SFML/SFML-Website). We're currently using MkDocs with Material for MkDocs as theme.

I don't recommend the way we currently integrate the SFML documentation, but if you go through the hoops of creating a layout that works, it can still be considered. I wanted to keep the SFML docs integrated as close as we had it before, for CSFML I can also see a different solution with the available Doxygen MkDocs plugins.

Sigmanificient:
Hmm, no sure how i could do it, i used sphinx and generated the docs quite aggressively through Makefile, with the doxygen xml behind the scene :sweat_smile:

PR #415, opened 2025-09-18
Update SFML version to 3.0.2

eXpl0it3r:
Making sure we're using the latest SFML 3.0 patch version


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ endif()
 if(CSFML_BUILD_NETWORK)
     list(PREPEND SFML_MODULES "Network")
 endif()
-find_package(SFML 3.0.0 COMPONENTS ${SFML_MODULES} REQUIRED)
+find_package(SFML 3.0.2 COMPONENTS ${SFML_MODULES} REQUIRED)
```
ChrisThrasher, line 69: Does CSFML need anything from 3.0.2 in particular? My understanding is that CSFML should work with any version of SFML 3.

eXpl0it3r: I guess we could stick to 3.0 or does 3.0.0 also allow to find newer patch versions? (Didn't I ask this in the past already? 😄)

eXpl0it3r: I think `3.0` makes the most sense, since we do want to target SFML 3.0 and don't care whether it's 3.0.0 or 3.0.2

